PR #50, opened 2026-02-09
Claude/focused dewdney

jbearak:
## Summary by CodeRabbit

* **Documentation**
  * Restructured README into an "Editor Setup" section with cross-editor LSP guidance, stdio usage, clarified install/build flow, expanded Quick Start and "When to Use" to emphasize cross-file symbol resolution and diagnostics, and condensed deprecated extension-specific guidance.

* **New Features**
  * Added redundant-directive severity setting and expanded cross-file severity options to accept "off" so each diagnostic category can be disabled.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,9 +16,9 @@ Pre-built binaries are available from the [releases page](../../releases).
 
 ### Editor Setup
 
-Raven is a standard Language Server Protocol (LSP) server. This repository includes a VS Code extension, but the binary works with any editor that supports LSP.
+Raven provides cross-file symbol resolution, position-aware scope, and diagnostics for R codebases. It works with any editor that supports the Language Server Protocol (LSP).
 
-**VS Code:** The extension is bundled with the binary. After running `setup.sh`, reload VS Code to activate.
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -14,8 +14,25 @@ cd raven
 ### Download from Releases
 Pre-built binaries are available from the [releases page](../../releases).
 
-### VS Code Extension
-The extension is bundled with the binary. After running `setup.sh`, reload VS Code to activate.
+### Editor Setup
+
+Raven is a standard Language Server Protocol (LSP) server. This repository includes a VS Code extension, but the binary works with any editor that supports LSP.
+
+**VS Code:** The extension is bundled with the binary. After running `setup.sh`, reload VS Code to activate.
```
jbearak: `setup.sh` is for development, so I wouldn't suggest running setup.sh. I also wouldn't say the extension is "bundled with" the binary. Rather, the _extension_ is what bundles the binary -- so that one can install the .vsix w/o also installing the binary. However, rather than explaining this, I think we might want to create accuracy by _deleting_ text. If anything, we could state that both (the extension and the binary) can be found on the Releases page (link)

PR #113, opened 2026-04-26
fix: log warning instead of unwrap on spawn_blocking join failure in debounce_loop

### LINKED ISSUE #111: Log warning on spawn_blocking JoinError in debounce_loop drain
## Context

In `crates/raven/src/libpath_watcher.rs` line ~494, the `spawn_blocking` call that drains queued notify events uses `.unwrap_or_default()` on `JoinError`:

```rust
let drained_paths: Vec<PathBuf> = tokio::task::spawn_blocking(move || {
    let mut paths = Vec::new();
    let guard = rx_arc.lock().unwrap();
    while let Ok(res) = guard.try_recv() { ... }
    paths
})
.await
.unwrap_or_default();
```

A `JoinError` here means the blocking task panicked (likely from a poisoned `StdMutex`) or the runtime is shutting down. The current code silently swallows this and continues with an empty path list, which means `touched_from_events` will miss in-place package upgrades for that debounce cycle.

## Suggested fix

Replace `.unwrap_or_default()` with an explicit match that logs a warning before falling back to an empty vec:

```rust
let drained_paths: Vec<PathBuf> = match tokio::task::spawn_blocking(move || {
    // ...
})
.await
{
    Ok(paths) => paths,
    Err(e) => {
        log::warn!("LibpathWatcher: drain task failed: {e}");
        Vec::new()
    }
};
```

This keeps the soft-failure behavior (losing drain paths is non-fatal — the diff still runs) while making the failure visible in logs for debugging.

## Why this is non-fatal vs the capture site

The other two `spawn_blocking` sites (`LibpathSnapshot::capture`) treat `JoinError` as fatal and send `LibpathEvent::Dropped`. That is correct because without a snapshot, no diff can run. The drain failure is different: the diff still runs against the new snapshot, it just has a smaller `touched` set for that one cycle. Logging is sufficient.

## References

- `crates/raven/src/libpath_watcher.rs` `debounce_loop` (~line 494)
- Identified during code review of #108

jbearak:
Replace `.unwrap_or_default()` on the `spawn_blocking` join handle with an
explicit match so a panicking drain task produces a visible warning instead of
silently dropping the error.

---

<a href="https://app.devin.ai/review/jbearak/raven/pull/113" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

Closes #111 

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced error handling and logging for internal event tracking to better diagnose processing failures when they occur.

### DIFF
--- a/crates/raven/src/libpath_watcher.rs
+++ b/crates/raven/src/libpath_watcher.rs
@@ -471,7 +471,7 @@ async fn debounce_loop(
                 };
                 tokio::time::sleep(debounce).await;
                 let rx_arc = Arc::clone(&raw_rx);
-                let drained_paths: Vec<PathBuf> = tokio::task::spawn_blocking(move || {
+                let drained_paths: Vec<PathBuf> = match tokio::task::spawn_blocking(move || {
                     let mut paths = Vec::new();
                     let guard = rx_arc.lock().unwrap();
                     while let Ok(res) = guard.try_recv() {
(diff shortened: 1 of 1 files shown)


jbearak:
@coderabbit review

PR #119, opened 2026-04-27
Fix package diagnostics for uninstalled packages and cold start

jbearak:
## Summary

Fixes a tangle of related bugs around the \`Undefined variable\` and \`Package not installed\` diagnostics for files containing \`library(P); P_func()\` when \`P\` is not installed. Symptoms ranged from "diagnostic flashes briefly then disappears" to "diagnostic never appears until the file is edited."

The fixes split cleanly across the diagnostic pipeline (suppression gates, package existence checks, scan-window handling, cross-file inheritance, force-republish gating). Each issue has its own commit so any future regression bisects to a specific fix.

## What changed (oldest → newest)

1. **\`4c1cf67\`** — \`package_cache_pending\` suppression must check \`package_exists()\` in addition to \`!is_cached_sync()\`. A never-installed package is also never cached, so without the existence check, uninstalled packages were permanently suppressed as if still loading.

2. **\`3f6412f\`** — "Package not installed" gate must not use \`r_subprocess.is_none()\` alone. \`ensure_package_library_initialized()\` can populate \`lib_paths\` without attaching an R subprocess; the correct gate is \`lib_paths().is_empty() && r_subprocess().is_none()\` (suppress only when there's no information at all).

3. **\`39e90a5\`** — \`PackageLibrary::package_exists()\` must consult the filesystem only, never the in-memory cache. \`prefetch_packages()\` inserts an empty-exports cache entry for any package R fails to load (asNamespace error swallowed by \`tryCatch\`), so trusting the cache made never-installed packages appear "installed."

4. **\`711b332\`** (later replaced by \`1565f15\`) — Initial attempt at per-identifier deferral during the workspace-scan window. Surfaced package-related function calls early but produced false positives for parent-defined helpers (\`library(dplyr); helper_fn()\` where \`helper_fn\` was actually defined in a sibling file).

5. **\`8d0fd12\`** — \`state.workspace_imports\` now stores \`(package, symbol)\` pairs (was \`Vec<String>\` of bare symbol names). The undefined-variable suppression that consults it must verify the source package is actually installed via \`package_library.package_exists(pkg)\` before suppressing — otherwise a workspace package whose NAMESPACE imports \`lmer\` from \`lme4\` silenced "Undefined variable: lmer" diagnostics workspace-wide even when \`lme4\` is not installed.

6. **\`745cf8f\`** — Post-workspace-scan path must call \`prefetch_packages_for_open_documents\` BEFORE force-republishing open documents. \`apply_workspace_index\` exposes auto-detected backward edges for the first time, so packages newly visible via inherited scope from now-indexed parent files were staying \`is_cached_sync = false\` forever (no later trigger cached them), and \`package_cache_pending\` permanently silenced \`lmer()\` after \`library(lme4)\` even though lme4 itself was uninstalled.

7. **\`1565f15\`** — Reverts \`711b332\`'s per-identifier deferral to a function-level early return. The textual \`has_prior_package_loader_call\` check was too coarse — any function call after any \`library(\` was treated as "package-related," producing flash-then-clear false positives for parent-defined helpers. Defer ALL undefined-variable diagnostics during the workspace-scan window; the post-scan force-republish re-runs the collector with the full dep graph in place.

8. **\`7378a74\`** — \`force_republish\` is now a counter (\`HashMap<Url, u32>\`) instead of a set. Each \`mark_force_republish\` increments; each \`record_publish\` decrements (saturating). With \`1565f15\` making the post-scan force-republish the only source of undefined-variable diagnostics for cold-opened files, a single intervening publish (e.g. the \`did_open\` stabilization task at T+750ms) was clearing the marker and blocking the post-scan revalidation at the same version, so the diagnostic was computed but never sent. Editing the file bumped the version and bypassed the gate, which is why diagnostics only appeared after a no-op space edit. The counter ensures N marks reliably get N publishes through the gate.

Plus: bumped \`@vscode/vsce\` to 3.9.1 (\`cd03cfd\`).

## Tests

- Added 7+ new unit tests covering each scenario: uninstalled packages, installed-but-uncached packages, missing-package gate with no R subprocess, \`package_exists\` ignoring cache, cold-start deferral, NAMESPACE import broken-by-uninstalled-source, post-scan prefetch from closed parent chain, multi-mark force-republish.
- Updated \`test_package_function_call_not_flagged_while_package_cache_pending\` and \`test_package_exists_cached_package_not_on_disk\` to reflect new semantics.
- All 2985 lib unit tests pass.

## Test plan

- [ ] Cold-open a file with \`library(uninstalled_pkg); some_fn()\` in a single-file workspace — both \"Package not installed\" and \"Undefined variable: some_fn\" should appear within ~1-2 seconds.
- [ ] Cold-open the same file inside a larger workspace where parent files in the source() chain load other (installed) packages — both diagnostics should still appear.
- [ ] Cold-open a file with \`library(dplyr); helper_fn()\` where \`helper_fn\` is defined in a sibling file — \`helper_fn\` should NOT be flagged once the workspace scan completes.
- [ ] Open a file that uses a NAMESPACE-imported symbol from a workspace package; uninstall the source package; diagnostic should appear after a refresh.
- [ ] Verify diagnostics appear on initial open without requiring a no-op edit.

---

<a href="https://app.devin.ai/review/jbearak/raven/pull/119" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **Bug Fixes**
  * More accurate diagnostics: "undefined variable" and "missing package" warnings are only suppressed when packages are confirmed installed on disk, preventing transient "appears then clears" behavior and ensuring deferral during workspace cold-start until scanning completes.
  * NAMESPACE-imported symbols no longer hide errors unless a mapped package is verified installed.
* **Refactor**
  * Forced republish handling now counts requests so multiple forced republish requests are honored independently.
* **Documentation**
  * Guidance updated to clarify caching and diagnostic-suppression rules.


jbearak:
### Code review

PR #122, opened 2026-04-27
fix: scope visibility for self-referential assignments

jbearak:
## Summary
- Emit "Undefined variable" diagnostic for `merp <- merp` (single-line and multi-line) — R evaluates the RHS *before* installing the LHS binding, so the LHS must not be in scope inside its own RHS.
- Decouple "anchor" and "visibility" positions on `ScopeEvent::Def`: `line/column` stays at the LHS for timeline sort and function-scope annotation; new `visible_from_line/visible_from_column` is the position at which the binding actually enters scope (end of assignment for non-function RHSs; LHS for function-definition RHSs, including parenthesized variants like `f <- (function() f())` so recursion still resolves).
- Sort the artifact timeline by each event's *effect* position via `event_effect_position`. This also fixes `x <- { rm(x); 1 }`, where the inner `rm(x)` was being applied *after* the Def event finalized the binding, silently dropping it.

Hover and go-to-definition are unaffected — the symbol's own `defined_line/defined_column` still points at the LHS identifier.

## Test plan
- [x] New test `test_self_referential_assignment_rhs_is_undefined` covers single-line and multi-line `merp <- merp`, plus the prior-binding case (`merp <- 1; merp <- merp` should not flag).
- [x] New test `test_assignment_def_processed_after_internal_rm` (and `test_assignment_binding_visible_after_rhs_with_internal_rm` at the diagnostic level) cover the timeline-ordering corollary.
- [x] New tests `test_parenthesized_function_rhs_visible_from_lhs` and `test_parenthesized_function_definition_rhs_supports_recursion` cover `f <- (function() f())` and `(function() g()) -> g`.
- [x] Existing scope, property, and integration tests updated to reflect that mid-assignment query positions correctly do not yet have the LHS binding installed.
- [x] Full suite passes: `cargo test -p raven` (2993 lib + 103 + 59 doctests).

## Notes for reviewers
- AGENTS.md gained four new Learnings: the two-position model, the test-position guidance, the timeline-sort invariant, and the paren-unwrap requirement.
- The diagnostic-level `merp <- merp` test is the user-facing regression check; the scope-level tests in `scope.rs` lock down the semantic invariants that the diagnostic test depends on (so a future change that papers over the bug at the diagnostic layer without fixing the scope semantics will still fail).

---

<a href="https://app.devin.ai/review/jbearak/raven/pull/122" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **Bug Fixes**
  * More accurate variable visibility timing so undefined-variable diagnostics appear only after an assignment's value becomes visible; timeline ordering updated so removals and definition finalization apply in correct effect/visibility order, fixing self-referential and parenthesized-recursive assignment cases.

* **Documentation**
  * Expanded scope-resolution guidance and diagnostic-publishing rules; formatting newline added.

* **Tests**
  * Added/refined tests for assignment visibility, rm()-style mutations, and recursive function forms.


jbearak:
### Code review

PR #130, opened 2026-04-29
perf: incremental scope resolution (parent-prefix cache + ScopeStream)

jbearak:
## Summary

Two-stage perf overhaul of cross-file scope resolution in the diagnostic path. End-to-end on `~/repos/worldwide`, the `data.r` post-scan diagnostic dropped from ~680 ms → ~35 ms (≈19.5×).

- **Stage 1: per-snapshot parent-prefix cache.** `scope_at_position_with_graph`'s STEP 1 (parent walk) is position-invariant for a queried URI within one `DiagnosticsSnapshot` — parameterized only by `query_inside_function`. Cache it in `ParentPrefixCache` keyed by `(Url, bool)`, owned by the snapshot. Same-file leak filters from earlier fixes run inside the cached region. ~680 ms → ~200 ms (3.4×).
- **Stage 2: ScopeStream for in-file timeline replay.** Diagnostic collectors that call `scope_at_position` per-identifier on the same file now walk usages in document order via a forward-only cursor (`crates/raven/src/cross_file/scope.rs`). Maintains a global frame plus a stack of active function-scope frames; pushes/pops on `FunctionScope` events; resolves each forward `source()` call site at most once and caches the contribution across strict and lazily-built late frames. Per-usage cost drops from O(all events ≤ usage) to O(events newly crossed). ~200 ms → ~35 ms (5.8×).
- **Parallel workspace scan.** `scan_workspace` splits into a serial directory walk (with symlink-cycle protection) and a `rayon::par_iter` per-file phase (read + parse + artifact compute) merged via fold/reduce. 413-file workspace: ~1.1 s → ~130 ms (8.5×).
- **DocumentStoreConfig defaults raised** to `max_documents: 4096`, `max_memory_bytes: 512 MB`. Original 50/100MB silently evicted open documents in workspaces with >50 R files, forcing `compute_artifacts_with_metadata` recomputation on every diagnostic pass and dwarfing every other suspect.
- **ScopeResolutionMemo** memoizes the recursive resolver across queries within a single diagnostic pass.
- Subtle correctness fixes captured along the way (all with regression coverage):
  - `event_effect_position` for `Source` and `Removal` events uses `column + 1` so the streaming path matches the recursive resolver's strict `<` comparison.
  - `ScopeStream` Source merges use `entry().or_insert()` (first-source-wins), matching the recursive resolver.
  - `advance_to`'s function-frame eviction uses `Vec::retain` over the full stack — sweeping past sibling functions used to leak the previous body's locals.
  - `assignment_rhs_is_function_definition` unwraps `parenthesized_expression` so `f <- (function() f())` still supports recursion.
  - `Def` events use end-of-assignment as `visible_from` for non-function RHS so `merp <- merp` resolves the RHS in the enclosing scope.

35 commits; +7,614 / −699 across `crates/raven/src/cross_file/scope.rs`, `handlers.rs`, `state.rs`, `backend.rs`, plan/perf-trace docs under `.claude/`, and proptest fixtures.

## Test plan
- [ ] `cargo test -p raven` (full suite incl. property tests + new regressions: `test_scope_stream_drops_sibling_function_frames`, `test_scope_stream_source_merge_precedence_matches_recursive`, `test_scope_stream_removal_effect_position_matches_recursive`, `test_self_referential_assignment_rhs_is_undefined`, `prop_scope_stream_matches_per_position`)
- [ ] `cargo bench --bench cross_file` and `cargo bench --bench edit_to_publish` — verify no regression
- [ ] Run `cargo run --example profile_real_workspace --features test-support` against a large R workspace (e.g. `~/repos/worldwide`) and confirm `data.r` post-scan diagnostic is ≤50 ms
- [ ] Smoke test in VS Code: open a multi-file R workspace, edit a file in a `source()` chain, confirm diagnostics update without flicker

---

<a href="https://app.devin.ai/review/jbearak/raven/pull/130" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Faster cold-start diagnostics via cached cross-file scope resolution and streaming evaluation.
  * Parallel workspace scanning for faster indexing.
  * New profiling examples that report end-to-end and per-file startup timings.

* **Bug Fixes**
  * Stronger package-name validation to avoid spurious background prefetch and subprocess work.
  * Improved out-of-scope/undefined-variable attribution accuracy.

* **Tests**
  * Deterministic 10-file self-ref fixture and additional regressions/property tests.

* **Documentation**
  * Expanded implementation plans and benchmarking reports.


jbearak:
@coderabbit review

jbearak:
### Code review

Found 1 issue:

1. `ScopeStream::is_visible` and `ScopeStream::symbol_for` iterate `prefix.symbols` linearly with `iter().any()` / `iter().find()` instead of using the O(1) `HashMap::contains_key` / `HashMap::get` that `Arc<str>: Borrow<str>` enables. CLAUDE.md says "Watch for O(n²) scope detection or duplicate-detection paths in large files; prefer indexed lookups when possible" and "lookups via `.get()` and `.contains_key()` need `&str` arguments... since `HashMap<Arc<str>, _>` implements `Borrow<str>`." Both methods are called per-identifier in the diagnostic hot loop (`collect_undefined_variables_from_snapshot`, `collect_out_of_scope_diagnostics_from_snapshot`), so for a file with thousands of usages and a parent prefix of hundreds of inherited symbols this produces O(usages × prefix_size) work — partially negating the streaming optimization on the workspaces this PR benchmarks against.

https://github.com/jbearak/raven/blob/62e9e44f07e94d88573bd26a1bfb32c91b313608/crates/raven/src/cross_file/scope.rs#L4282-L4285

https://github.com/jbearak/raven/blob/62e9e44f07e94d88573bd26a1bfb32c91b313608/crates/raven/src/cross_file/scope.rs#L4386-L4392

PR #153, opened 2026-05-03
Fix qualified-member goto-def across source chains

### LINKED ISSUE #148: qualified-member goto-def: scan cursor's full cross-file connected component, not just cursor + defining file
## Background

`crates/raven/src/qualified_resolve.rs` resolves go-to-definition for the RHS identifier of `$` and `@` (e.g. cmd-click on `bar` in `foo$bar`). It collects member-assignment / constructor candidates and picks the best one.

After the cross-file extension landed (PR with the `gotodef` branch — module-level doc comment in `qualified_resolve.rs` step 2), the resolver scans **two** files:

1. `foo`'s defining file.
2. The cursor's file (when distinct), with a per-site `get_cross_file_scope` check confirming `foo` at each candidate's LHS position resolves to the same binding.

This handles the common "skeleton-and-attach" pattern: `foo <- list()` in helpers.R, `foo\$bar <- 1` in main.R, cursor in main.R. ✅

## Gap

A `foo\$bar <- ...` in some **third** file in the middle of a `source()` chain is still missed. Concretely:

```r
# a.R
foo <- list()

# b.R
source(\"a.R\")
foo\$bar <- 1   # ← member assignment in a third file

# c.R  (cursor here)
source(\"b.R\")
print(foo\$bar)   # ← cmd-click on \`bar\` returns None
```

- `defining_uri` = a.R (where `foo` was bound).
- Cursor file = c.R.
- Scan covers a.R + c.R; b.R is invisible.
- a.R has no `bar` and c.R has no `foo\$bar <- ...` candidate, so the resolver returns `None`.

The user would expect goto-def on `\$bar` to jump to `foo\$bar <- 1` in b.R.

## Acceptance criteria

1. Extend candidate collection to walk **every file in the cursor's cross-file connected component** (forward + backward closures via the existing `state.cross_file_graph`), in addition to the cursor file and defining file.
2. Apply the same per-site \`get_cross_file_scope\` correctness gate to every non-defining-file candidate (including b.R candidates) — verify that \`foo\` at the candidate's LHS position resolves to the same `ScopedSymbol` the user navigated from.
3. Tie-break: existing rule continues — cursor-file candidates beat all others; among non-cursor-file candidates, max effect position wins. (Cross-file effect-position ordering is undefined in general; if it becomes an issue, prefer candidates whose file is closer to the cursor in the dependency graph.)
4. Add tests:
   - Positive: the three-file example above resolves to b.R's `foo\$bar <- 1`.
   - Negative: a `foo\$bar <- 99` inside an unrelated function in b.R (where `foo` is shadowed) does **not** match.
   - Negative: a `foo\$bar <- 99` in some unrelated file *outside* the cursor's connected component does not match.

## Notes

- The per-site \`candidate_lhs_matches_symbol\` check in qualified_resolve.rs already provides the correctness gate; the only missing piece is enumerating the right set of files to walk.
- The scope-resolve cost is bounded by the number of \`foo\$bar <- ...\` text matches across the connected component, which is small in practice.
- Documented as a known limitation in the module-level docstring of \`qualified_resolve.rs\`.

## Files to touch

- \`crates/raven/src/qualified_resolve.rs\` (\`resolve_qualified_member\`, the cursor-file phase) — generalize the file enumeration.
- \`crates/raven/src/state.rs\` / \`crates/raven/src/cross_file/\` — there's likely already a helper for \"reachable URIs from the cursor file\"; reuse it.

jbearak:
## Summary
- Scan the cursor file's full cross-file neighborhood for qualified-member assignments instead of only the cursor and defining files.
- Keep per-site scope validation so shadowed or unrelated `foo$bar <- ...` assignments are rejected.
- Update qualified-member design notes and regression tests for the three-file source chain.

## Tests
- `cargo fmt`
- `cargo test -p raven qualified_resolve`
- `cargo test -p raven`

## Artifacts
- Conversation: https://app.warp.dev/conversation/d9901a72-d31b-4858-9288-8e515411021d

Closes #148

## Summary by CodeRabbit

* **New Features**
  * More accurate go-to-definition for `$` and `@`, searching and preferring closer contributing files across the workspace.
  * Improved JAGS completion that excludes R-only identifiers.

* **Documentation**
  * Updated specs and guides describing the refined member-access/goto-definition behavior.

* **Tests**
  * Added regressions and updated property tests to validate the new resolution behavior.

### DIFF
--- a/crates/raven/proptest-regressions/handlers.txt
+++ b/crates/raven/proptest-regressions/handlers.txt
@@ -13,3 +13,4 @@ cc 2250c8f29bc61283775c429485d811bc470b61dc653492a4f3296a3aab75251d # shrinks to
 cc 983b057cd15b41227acf67a33b60f8aad42d25970dfa29401494b609f8d246fe # shrinks to reserved_word = "if", var_name = "ifa"
 cc d3a667aae8fefd13d4da291efa222936bb1f1f6806e9bf2c52529364e215c4de # shrinks to symbol_name = "bmp", is_function = false, lines_between = 0
 cc 45e6028c72f7d8b279bbde2cb156a3ac934a6cff2e4f83142171fa19e7706317 # shrinks to func = "i__", lhs = "aa0", rhs = "i__"
+cc 5ffd8e9d8fe0cb62ca747b14ce4eea20ff072c71b60bc3fcdeb7dd6869a03705 # shrinks to content = "NA<-", ext = "jags"
--- a/crates/raven/src/cross_file/scope.rs
+++ b/crates/raven/src/cross_file/scope.rs
@@ -743,6 +743,10 @@ impl Default for ScopeArtifacts {
 pub struct ScopeAtPosition {
     pub symbols: HashMap<Arc<str>, ScopedSymbol>,
     pub chain: Vec<Url>,
+    /// Widest query position from each contributing file that was visible when
+    /// this scope was computed. Forward-sourced files typically contribute at
+    /// EOF; backward-parent files contribute only up to their call site.
+    pub visible_positions: HashMap<Url, (u32, u32)>,
     /// URIs where max depth was exceeded, with the source call position (line, col)
     pub depth_exceeded: Vec<(Url, u32, u32)>,
     /// Packages inherited from parent files (loaded before the source() call site)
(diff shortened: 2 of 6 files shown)


jbearak:
### Code review

PR #157, opened 2026-05-03
Use cached scope for qualified member goto-def

### LINKED ISSUE #155: qualified-member goto-def: scope re-resolve loop holds WorldState read lock under N uncached scope walks
## Summary

After #153, `resolve_qualified_member` re-resolves cross-file scope once per `foo\$bar <- ...` site found anywhere in the cursor's connected component. On large workspaces (hundreds of connected files), or for common member-assignment patterns (e.g. `obj\$id`, `data\$name`), that loop can do many full uncached scope walks while the `WorldState` read lock is still held. Concurrent `did_change` writers wait for the entire loop to finish.

## Where

- Loop entry point: [`qualified_resolve.rs:212`](https://github.com/jbearak/raven/blob/6a7e15fa54b4390df8ced0e84dc80d351ae7a984/crates/raven/src/qualified_resolve.rs#L210-L215) — `cross_file_candidates.retain(|c| candidate_lhs_matches_symbol(state, c, lhs_name, symbol));`
- Per-candidate scope resolve: [`qualified_resolve.rs:228-246`](https://github.com/jbearak/raven/blob/6a7e15fa54b4390df8ced0e84dc80d351ae7a984/crates/raven/src/qualified_resolve.rs#L228-L246) — calls `get_cross_file_scope` with `DiagCancelToken::never()`.
- Uncached entry point: [`handlers.rs:2734`](https://github.com/jbearak/raven/blob/6a7e15fa54b4390df8ced0e84dc80d351ae7a984/crates/raven/src/handlers.rs#L2734-L2747) — calls `scope_at_position_with_graph` (no `ParentPrefixCache`).
- Read lock holder: [`backend.rs:3368`](https://github.com/jbearak/raven/blob/6a7e15fa54b4390df8ced0e84dc80d351ae7a984/crates/raven/src/backend.rs#L3364-L3374) — `goto_definition` holds `self.state.read().await` for the entire call.

A cached variant exists already: `scope_at_position_with_graph_cached` (`scope.rs:2926`), which memoizes the parent-prefix walk via `ParentPrefixCache`.

## Impact

- Each retain iteration walks the dependency graph and merges artifacts from scratch — no prefix reuse across iterations, even when many candidates share a file.
- The contract on `get_cross_file_scope` ([`handlers.rs:2683-2693`](https://github.com/jbearak/raven/blob/6a7e15fa54b4390df8ced0e84dc80d351ae7a984/crates/raven/src/handlers.rs#L2683-L2693)) describes interactive handlers as resolving \"a single position per request.\" That framing is no longer accurate for `resolve_qualified_member`; it now resolves up to one position per matching site across the neighborhood.
- The cancel token is `DiagCancelToken::never()`, so even a stale request can't bow out early.

## Proposed mitigations (any one helps, combining is best)

1. **Use the cached entry point.** Build a `ParentPrefixCache` at the start of `resolve_qualified_member` and route both the initial `foo` resolve and the per-candidate re-resolve through `scope_at_position_with_graph_cached`. Most savings come from shared parent prefixes across same-file candidates and across the cursor file / defining file / neighborhood files.
2. **Group candidates by URI before resolving.** The per-site validation only needs `lhs_name → ScopedSymbol` equality at the candidate's LHS position. Within one file, the parent prefix is position-invariant, so all candidates in that file can share a single prefix-walk's worth of work; only the inner-scope step is per-position.
3. **Pass a request-scoped cancel token** so the loop bails out when the user moves the cursor or the request is superseded.

## Repro hook

Open a workspace with ~200 sourced R files where a common identifier (e.g. `state\$x`) has many member-assignments throughout the chain, then trigger goto-def on `\$x` while another file is being edited. Compare time-under-read-lock before/after #153.

## Notes

- Not a correctness regression — only a latency / writer-starvation concern.
- The locking discipline rule in CLAUDE.md narrowly targets diagnostics and the libpath consumer's filter, so this isn't a strict violation, but it's the same anti-pattern (iterate-many-scope-resolves-under-the-read-lock) at a smaller scale.

jbearak:
## Summary
- Use a request-local `ParentPrefixCache` for qualified-member goto-definition scope lookups.
- Add a cached cross-file scope helper for interactive requests that resolve multiple related positions.
- Update scope helper locking guidance for multi-position interactive paths.

## Validation
- `cargo fmt --check`
- `cargo test -p raven qualified_resolve::tests --lib`
- `cargo test -p raven`

Fixes #155.

Oz conversation: https://app.warp.dev/conversation/a0494f9a-1533-4416-b3d2-7a1d385cb0ae

## Summary by CodeRabbit

* **Documentation**
  * Clarified documentation for snapshot-scoped cache usage patterns in cross-file scope resolution, including instantiation guidelines and lifetime examples for both diagnostic passes and interactive request scenarios.

* **Refactor**
  * Improved performance for multi-position request handling by implementing an optimized caching mechanism for scope resolution, reducing redundant computations across interactive request batching and diagnostic analysis phases.

### DIFF
--- a/crates/raven/src/cross_file/scope.rs
+++ b/crates/raven/src/cross_file/scope.rs
@@ -2924,9 +2924,17 @@ where
 /// Per-snapshot cache for `ParentPrefix` results, keyed by `(target URI,
 /// query_inside_function)`.
 ///
-/// Lives inside `DiagnosticsSnapshot` so it shares the snapshot's lifetime;
-/// never shared across snapshots. Within one diagnostic pass, all scope queries
-/// against the same URI reuse the same prefix entries.
+/// Use one instance per `WorldState` snapshot and never share across
+/// snapshots. Two patterns:
+///
+/// - **Diagnostic pass**: lives inside `DiagnosticsSnapshot` so it shares the
+///   snapshot's lifetime; all scope queries against the same URI reuse the
+///   same prefix entries.
+/// - **Interactive request batch**: a request handler that resolves many
+///   positions under one read guard (for example
+///   `qualified_resolve::resolve_qualified_member`) creates a cache at the
+///   start of the call, threads it through `get_cross_file_scope_with_cache`,
+///   and drops it when the call returns.
 #[derive(Debug, Default)]
 pub struct ParentPrefixCache {
     entries: HashMap<(Url, bool), Arc<ParentPrefix>>,
--- a/crates/raven/src/handlers.rs
+++ b/crates/raven/src/handlers.rs
@@ -2706,34 +2706,17 @@ pub(crate) fn get_cross_file_scope(
     column: u32,
     cancel: &DiagCancelToken,
 ) -> scope::ScopeAtPosition {
-    // Use ContentProvider for unified access
     let content_provider = state.content_provider();
-
-    // Closure to get artifacts for a URI
     let get_artifacts = |target_uri: &Url| -> Option<Arc<scope::ScopeArtifacts>> {
         content_provider.get_artifacts(target_uri)
     };
-
-    // Closure to get metadata for a URI
     let get_metadata =
         |target_uri: &Url| -> Option<std::sync::Arc<crate::cross_file::CrossFileMetadata>> {
             content_provider.get_metadata(target_uri)
         };
-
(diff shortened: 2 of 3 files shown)


jbearak:
@coderabbit review

PR #164, opened 2026-05-05
Exclude parent-prefix member assignments from child completions

jbearak:
Track URIs that enter the scope chain via the parent/backward prefix in
ScopeAtPosition::parent_prefix_chain_uris. Filter these out when
collecting qualified member candidates, since parent files execute after
the child and their member assignments (e.g. foo$bar <- 1) are not
visible at the child's cursor position.

Adds test to verify parent-prefix assignments don't leak into child
completions.

## Summary by CodeRabbit

* **Tests**
  * Added regression tests to ensure member completion and goto-definition behavior correctly respects file dependency boundaries across sourced files.
  * Verifies that member assignments from parent files are only visible to dependent files when defined before the source statement, ensuring proper variable scoping.


jbearak:
@coderabbit review

jbearak:
@Coderabbit review

PR #169, opened 2026-05-05
Fix parent-after-source leak in cross-file member completions

jbearak:
## Summary

In a workspace with a *diamond* of `source()` relationships, `ww$c.oos` and `ww$seeds` assigned in `main.r` *after* `source("scripts/data.r")` were appearing in member completions inside `data.r` — even though they execute strictly after `data.r` returns and shouldn't be in scope yet.

Root cause: when computing `data.r`'s `parent_prefix`, a transitive walk visits a grandparent of `main.r` at a *wider* position than its initial call-site visit (because that grandparent also sources a sibling forward-sourced from `main.r`). The re-visit's STEP 2 forward-sources `main.r` at EOF, and `record_visible_position` overwrote `main.r`'s cap from `(call_site, 0)` to `(MAX, MAX)`. That EOF cap propagated back through the merge chain, defeating the visibility filter in `qualified_resolve::candidate_effect_visible_in_scope`.

Fix: pin the cap on re-visit. `record_visible_position` now runs only on the first visit. The first visit's position is the legitimate cap for that file's role in the cursor's execution chain — side-trip recursions must not widen it. STEP 2 still re-runs at the wider position to collect any new symbols; only the visibility cutoff stays put.

- Reproduces against `worldwide/scripts/data.r` (`ww$|` → `c.oos`, `seeds` no longer appear).
- New regression test exercises the minimal diamond.
- All 3211 lib tests pass.

## Test plan

- [x] `cargo test -p raven --lib qualified_resolve::tests::dollar_member_completion_excludes_parent_after_source_via_transitive_revisit` — passes.
- [x] TDD red→green verified: revert the fix → test fails with `["c.oos", "name.w", "seeds"]`; restore → test passes with `["name.w"]`.
- [x] `cargo test -p raven --lib` — 3211 passed, 0 failed.
- [x] Manual completion request against the actual `~/repos/worldwide` repo at `scripts/data.r` line 28 — completion list goes from 75 labels (with `c.oos`, `seeds`) to 73 labels (without).

## Summary by CodeRabbit

* **Bug Fixes**
  * Cross-file scope resolution now records symbol visibility only on the initial visit, preventing unintended expansion of visibility during revisits and ensuring correct handling of cycles and forward-source anchoring.
  * Prevented parent-prefix visibility leakage in member completion across transitive file relationships, preserving proper scope boundaries.

* **Tests**
  * Added test coverage for dollar-member completion behavior across transitive file dependencies.

### DIFF
--- a/crates/raven/src/cross_file/scope.rs
+++ b/crates/raven/src/cross_file/scope.rs
@@ -3438,13 +3438,16 @@ where
 ///
 /// # Invariants
 ///
-/// **`visible_positions[uri]` is pinned to the first-visit position.** A re-visit
-/// happens when a transitive cycle forward-sources `uri` at a wider position
-/// (typically EOF). The *original* visit already recorded the legitimate cap for
-/// `uri`'s role in the cursor's execution chain; expanding it on re-visit lets a
-/// side-trip recursion overwrite the cap and reveal later-than-source definitions
-/// that aren't actually in scope. The re-visit still re-runs STEP 2 to collect any
-/// new symbols — only the visibility cutoff stays put.
+/// **`visible_positions[uri]` is pinned to the first-visit position.** A
+/// re-visit happens when the recursion re-enters `uri` at a strictly wider
+/// position than the first-visit cap (typically a transitive forward
+/// `source()` reaching `uri` at EOF). The *original* visit already recorded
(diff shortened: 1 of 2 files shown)


jbearak:
@coderabbit review

jbearak:
@CodeRabbit review

PR #251, opened 2026-05-15
refactor(scaffold): drop hand-rolled JSONC walkers in favour of jsonc-parser

### LINKED ISSUE #250: scaffold: revisit hand-rolled JSONC walkers (findOutermostClosingBrace, stripJsoncComments) if we add JSON5 features
Follow-up from #249. None of these are blocking — flagging for future cleanup so we don't lose track.

## Context

The `Raven: Create linting settings` scaffold (`editors/vscode/src/scaffold.ts`) uses [`microsoft/jsonc-parser`](https://github.com/microsoft/node-jsonc-parser) for the JSONC heavy lifting: `parseTree` for structure analysis (top-level keys, value types, parse errors), `visit` for sentinel comment detection, and `modify` + `applyEdits` for removing existing `raven.linting.*` keys. The only remaining hand-rolled JSONC helper is `hasCommaBetween` — a ~15-line comment-aware scan that decides whether the last existing top-level property already has a trailing comma before our block goes in. We use it instead of `modify`-for-insertion because `modify`'s comma logic also migrates the user's trailing `// inline comment` onto the new key's line, which is surprising UX.

## Known limitations

- **One-key-per-line assumption** in the comma-decision helper. If a user puts `"editor.tabSize": 4, "files.autoSave": "onFocusChange"` on the same line, our scan still works (it stops at the first `,`), but for any *new* hand-rolled JSONC code in this file the assumption needs to stay documented. The function's doc-comment links here.
- **No JSON5 number/string literals** (single quotes, trailing-`.` decimals, unquoted keys). `jsonc-parser` rejects these as parse errors and the scaffold refuses to merge — clean failure, but not the friendliest UX. The user has to convert to plain JSONC and re-run.

## Already addressed in PR #249

For posterity, so future readers don't dig back through the review thread:

- ✅ Re-runs duplicating the `// lintr: …` header block — fixed via sentinel-wrapped block (`// >>> raven.linting.* (managed by ...)` / `// <<<`) + sentinel-aware stripping.
- ✅ Comma inserted after a trailing `//` comment landing inside the comment — fixed: the comma goes right after the last property's value, before the inline comment.
- ✅ Nested `raven.linting.*` keys inside `[r]` overrides silently stripped — fixed: `jsonc-parser`'s `parseTree` walks only top-level properties, and `modify` removes by path (`[keyName]`) which targets only top-level keys.
- ✅ Non-object root JSON (e.g. `[]`) silently clobbered — fixed via `nonObjectRoot` discriminator in `classifyExisting`.
- ✅ Sentinel-shaped lines inside a `/* ... */` block comment treated as real sentinels — fixed: `visit.onComment` fires once per *whole* comment, so embedded `//`-shaped text inside a block comment is not separately reported.
- ✅ Unterminated `/*` block comment swallowing rest of file — fixed: `jsonc-parser` reports the unterminated comment as a parse error.
- ✅ Non-scalar `raven.linting.*` values orphaning continuation lines — fixed via `unsupportedValue` discriminator in `classifyExisting`.
- ✅ Hand-rolled JSONC walkers (`stripJsoncComments`, `findOutermostClosingBrace`, `findSentinelLineRange`, `stripTopLevelLintingLines`, `appendCommaIfNeeded`) duplicating string/comment state machines — replaced with `jsonc-parser` (~250-line net reduction in `scaffold.ts`).

## Nice-to-have cleanups

- The `// >>> raven.linting.* (managed by "Raven: Create linting settings")` sentinel is long. If we ever rename the command, the sentinel needs to stay backwards-compatible (so a re-run can still find the old block from prior versions). Worth a small migration matrix when that happens.
- The build now sets `mainFields: ['module', 'main']` in `editors/vscode/scripts/build.js` so esbuild bundles `jsonc-parser`'s ESM build (its UMD wrapper's dynamic `require("./impl/...")` calls don't survive esbuild's static analysis). If we ever add a Node-platform dep that explicitly *needs* its UMD/CJS build, we'd need to special-case it.

## Acceptance

Close this issue if (a) we rename the scaffold command (triggering a sentinel migration), or (b) a year passes without anyone hitting these limitations.

jbearak:
Follow-up to #249 (merged at `bb6c0f6`). Addresses the hand-rolled JSONC walkers tracked in #250.

## Summary

- Adopt [`microsoft/jsonc-parser`](https://github.com/microsoft/node-jsonc-parser) (MIT, ~30 KB) as a direct runtime dependency.
- Drop the five hand-rolled walkers (`stripJsoncComments`, `findOutermostClosingBrace`, `findSentinelLineRange`, `stripTopLevelLintingLines`, `appendCommaIfNeeded`) — they each carried their own copy of the string/comment/escape state machine, and the three rounds of Codex review on #249 caught a separate drift in each. Net `scaffold.ts` is **~250 lines smaller** (460 removed / 229 added).
- New code uses `parseTree` for structure analysis (top-level keys, value types, parse errors), `visit.onComment` for sentinel detection (which inherently ignores sentinel-shaped text inside block comments or strings), and `modify` + `applyEdits` to remove existing `raven.linting.*` keys (including any inline `// comments` attached to them).
- Build tweak: `editors/vscode/scripts/build.js` now sets `mainFields: ['module', 'main']` so esbuild bundles `jsonc-parser`'s ESM build. Without this, the UMD wrapper's dynamic `require("./impl/...")` calls don't survive bundling and the extension crashes at activation with `Cannot find module './impl/format'`.

## What stayed hand-rolled and why

A small `hasCommaBetween` helper (~15 lines, comment-aware) decides whether the last existing top-level property already has a trailing comma. `jsonc-parser`'s `modify`-for-insertion would handle that automatically — but it also migrates the user's trailing `// inline comment` from the last property onto the new key's line. Preserving the user's inline comments where they wrote them is better UX, so this helper takes that decision back.

## Why this didn't land with #249

The user merged #249 at `bb6c0f6` while I was finishing this refactor (commit `cbb17c5` on the now-merged branch). Rather than reopening, this stands as a clean follow-up against the merged state of main.

## Test plan

- [x] `bun run typecheck` (vscode/)
- [x] `bun run test` — 201 mocha tests pass, including the sentinel / non-scalar / non-object / parse-error / unterminated-comment / block-comment-sentinel / nested-`[r]`-override / inline-comment-preservation cases
- [x] Full `bun test` workspace suite (cargo + vscode-mocha): 467/467 green
- [x] Codex review run on the refactor: "All five concerns clear — no action required"
- [ ] Manual smoke: run "Raven: Create linting settings" on a fresh workspace, on a workspace whose `.vscode/settings.json` has unrelated keys, and on one with a pre-existing scaffold block (re-run should be idempotent and silent).

Closes #250.

---

<a href="https://app.devin.ai/review/jbearak/raven/pull/251" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **Refactor**
  * More reliable linting settings scaffolding: safer, comment-aware JSONC parsing and merges that better preserve user-authored settings while inserting managed linting config.
  * Build adjustments to ensure consistent module resolution during packaging.

* **Tests**
  * Added comprehensive tests covering JSONC edge cases for scaffolded linting merges.

* **Chores**
  * Pinned a JSONC parser dependency to improve parsing stability.

[Review Change Stack](https://app.coderabbit.ai/change-stack/jbearak/raven/pull/251?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/editors/vscode/package-lock.json
+++ b/editors/vscode/package-lock.json
@@ -10,6 +10,7 @@
       "license": "GPL-3.0",
       "dependencies": {
         "apache-arrow": "^21.1.0",
+        "jsonc-parser": "^3.3.1",
         "vscode-languageclient": "^9.0.1"
       },
       "devDependencies": {
@@ -4231,7 +4232,6 @@
       "version": "3.3.1",
       "resolved": "https://registry.npmjs.org/jsonc-parser/-/jsonc-parser-3.3.1.tgz",
       "integrity": "sha512-HUgH65KyejrUFPvHFPbqOY0rsFip3Bo5wb4ngvdi1EpCYWUQDC5V+Y7mZws+DLkr4M//zQJoanu1SP+87Dv1oQ==",
-      "dev": true,
       "license": "MIT"
     },
     "node_modules/jsonfile": {
--- a/editors/vscode/package.json
+++ b/editors/vscode/package.json
@@ -1632,6 +1632,7 @@
   },
   "dependencies": {
     "apache-arrow": "^21.1.0",
+    "jsonc-parser": "^3.3.1",
     "vscode-languageclient": "^9.0.1"
   },
   "overrides": {
--- a/editors/vscode/scripts/build.js
+++ b/editors/vscode/scripts/build.js
@@ -9,6 +9,18 @@ const sveltePreprocess = require('svelte-preprocess').default ?? require('svelte
 const root = path.resolve(__dirname, '..');
 const dist = path.join(root, 'dist');
 
+// `jsonc-parser` ships both a UMD `main` (`./lib/umd/main.js`) and an ESM
+// `module` (`./lib/esm/main.js`). esbuild's Node platform defaults to
+// `mainFields = ['main']`, which picks the UMD build — whose wrapper uses
+// dynamic `require("./impl/format")` / `./impl/edit` / etc. calls that
+// esbuild can't statically follow. The bundle compiles but throws
+// "Cannot find module './impl/format'" at activation time. Aliasing
+// `jsonc-parser` directly to its ESM entry gives esbuild plain `import`
+// statements it can inline, and keeps the fix scoped — resolution for
+// every other runtime dependency (apache-arrow, vscode-languageclient,
+// …) stays on its default `main` field.
+const jsoncParserEsmEntry = require.resolve('jsonc-parser/lib/esm/main.js');
+
 async function buildExtension() {
     await esbuild.build({
         entryPoints: [path.join(root, 'src', 'extension.ts')],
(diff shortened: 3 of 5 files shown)


jbearak:
### Code review

PR #255, opened 2026-05-16
feat(diagnostics): flag invalid assignment targets (#34)

### LINKED ISSUE #34: Diagnostics for invalid assignment targets
- `else <- 1` — assigning to a reserved word (`else`, `if`, `for`, `while`, `in`, `function`, `repeat`, `next`, `break`)
- `TRUE <- 5` — assigning to a logical literal
- `NULL <- "x"` — assigning to `NULL`
- `1 <- "x"` — assigning to a numeric literal
- `"foo" <- 1` — assigning to a string literal
- Applies to all assignment operators: `<-`, `=`, `<<-`, `->`

Note: R actually allows some of these at the REPL (e.g., `T <- FALSE` works because `T` is not a reserved word, just a default binding). The diagnostics should target cases that are truly invalid syntax or that tree-sitter reports as `ERROR` nodes.

jbearak:
## Summary

Closes #34. Adds a built-in error diagnostic for assignments whose target is something R will always reject — literals (`TRUE`, `FALSE`, `NULL`, any `NA*`, `Inf`, `NaN`, numeric, string) or reserved-word identifiers tree-sitter still parses as `binary_operator` (`else <- 1`, `in <- 1`, `next <- 1`, `break <- 1`).

Implemented as a built-in collector in `handlers.rs` (next to `collect_syntax_errors` / `collect_else_newline_errors`), **not** a lint rule: these are R errors, not stylistic preferences, and shouldn't be hidden by the opt-in lint master switch.

- Covers all five assignment operators: `<-`, `<<-`, `=`, `->`, `->>`. For right-arrow operators the target is the RHS.
- Skips named-argument `=` (`f(x = 1)`) and default-value `=` in formal parameters (`function(x = TRUE) ...`) — tree-sitter wraps those in `argument` / parameter nodes, so they're naturally outside the check.
- Skips cases where tree-sitter already emits an `ERROR` node (`if <- 1`, `for <- 1`, `while <- 1`, `function <- 1`) so `collect_syntax_errors` remains the single source of one diagnostic per site.
- Honors the issue note: `T <- FALSE` / `F <- TRUE` are **not** flagged. `T` and `F` are ordinary bindings that default to `TRUE`/`FALSE`; R accepts the rebind. The style-lint `T`/`F` rule still flags bare `T` / `F` in reference position.

## Test plan

- [x] `cargo test -p raven --lib` (3735 passed; 0 failed) — includes 24 new tests in `invalid_assignment_target_tests`
- [x] `cargo test --release -p raven --features test-support` (same as CI integration job; all green)
- [x] Probed tree-sitter-r's actual AST for every case in the issue to make sure the predicate covers the right node kinds (literals → `true`/`false`/`null`/`na`/`inf`/`nan`/`float`/`integer`/`complex`/`string`; `next`/`break` → own kinds; `else`/`in` → `identifier` + reserved-word check; `if`/`for`/`while`/`function` → `ERROR`, deferred to `collect_syntax_errors`).
- [x] Range correctness: diagnostic points at the offending target (not the operator); UTF-16 column math is exercised by a test with a non-ASCII prefix.
- [ ] CI: integration tests + vscode-mocha to run on push.

---

<a href="https://app.devin.ai/review/jbearak/raven/pull/255" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added diagnostics for invalid assignment targets (assigning to literals, NA/Inf/NaN, signed numeric literals, reserved words) and warnings for suspicious targets (string-literal names, ...), across all assignment operators; diagnostics highlight the target token and honor inline suppression comments.
* **Documentation**
  * Documented the new "Assignment Targets" diagnostics and their scope/exceptions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/jbearak/raven/pull/255?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/crates/raven/src/handlers.rs
+++ b/crates/raven/src/handlers.rs
@@ -7754,13 +7754,22 @@ mod invalid_assignment_target_tests {
 
     #[track_caller]
     fn assert_one(code: &str, expect_substr: &str) {
+        assert_one_with_severity(code, DiagnosticSeverity::ERROR, expect_substr);
+    }
+
+    #[track_caller]
+    fn assert_one_with_severity(
+        code: &str,
+        expected: DiagnosticSeverity,
+        expect_substr: &str,
+    ) {
         let diags = collect(code);
         assert_eq!(
             diags.len(),
             1,
             "expected exactly one diagnostic for `{code}`, got {diags:?}"
         );
-        assert_eq!(diags[0].severity, Some(DiagnosticSeverity::ERROR));
+        assert_eq!(diags[0].severity, Some(expected));
         assert!(
             diags[0].message.contains(expect_substr),
             "message `{}` did not contain `{}`",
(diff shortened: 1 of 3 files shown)


jbearak:
### Code review

Reviewed five angles in parallel — CLAUDE.md adherence, shallow bug scan, git history regression check, prior-PR comment patterns, and code-comment guidance. Several documentation nits surfaced but all scored below the high-confidence threshold; they've been addressed in [136ccb7](https://github.com/jbearak/raven/commit/136ccb7293aa1c5cd5c86da2727e1797a7fb1056) (doc: list `repeat` alongside the other ERROR-parented control-flow keywords and note that both severity tiers honor `# @lsp-ignore`; test: replace the BMP `ø` with `🦀` so the UTF-16 column-math test actually exercises the surrogate pair).

PR #326, opened 2026-05-26
Data viewer: sort rows by column(s)

### LINKED ISSUE #322: Data Viewer: Sort rows by column(s)

jbearak:
## Summary

Closes #322.

Sort the data viewer's rows by one or more columns. Right-click a column header to apply, append, flip, or clear; sorted headers show an arrow plus a 1-based priority badge when multi-key; a toolbar chip strip lists active keys with per-chip Flip / Remove / Move-to-first.

- Sort runs in the extension host against the on-disk Arrow file: each sort-key column is read once, then a `Uint32Array` permutation is built with a stable, NA-last lex comparator. `getRows` accepts an optional permutation and serves rows in visible order grouped by batch (so the batch LRU stays warm). Copy follows the same path — copied TSV matches the visible order.
- Sort key derivation is WYSIWYG: factor and value-labelled columns sort by the displayed label when Labels is on and by the underlying code when off; toggling Labels with a sort active on a labelled column re-sorts. Format never affects order.
- Sort state persists per panel-name × schema hash alongside the existing layout and toolbar stores. Restored on a later `View()` against the same shape; dropped silently if `nrow` differs. Behind `raven.dataViewer.persistSort` (default `true`).
- Keyboard: `Shift+Alt+A` / `Shift+Alt+D` sort the focused column; `Shift+Alt+0` clears.

## Test plan

- [x] `cargo build -p raven` and the bun suite both green (1175 passing — 42 new across sort engine, permuted reader, sort-state store, and a panel-level setSort/sortApplied/saveSort/restore round-trip)
- [x] TypeScript typecheck clean across all four tsconfigs
- [ ] Manual: open `View(mtcars)`, right-click `mpg` → Sort ascending, verify arrow + status-bar summary + chip strip
- [ ] Manual: shift-click `cyl` → Sort descending (or use the new *Add descending to sort* item), verify priority badge "2" appears on the second sorted header
- [ ] Manual: click the `mpg` chip → Flip direction; click again → Remove from sort; click ✕ → Clear all
- [ ] Manual: with a sort active, run `View(mtcars)` a second time and confirm the sort restores
- [ ] Manual: sort a factor or `haven_labelled` column, toggle Labels off/on, confirm the rows re-sort

## Notes

- The header pulse-on-apply animation described in the spec is deferred; the chip strip and header glyphs make the sort state plain enough for v1.
- glide-data-grid v6's `rowMarkers` is offset-only, so the row-number gutter still shows visible position rather than the original row index. The original index round-trips on the `rows` wire reply (`originalRowIndices`); we'll surface it when the marker API allows per-row content.
- Worker-thread offload for ≥500k-row sorts isn't wired yet; the engine is correct on the main event loop and the LRU plus typed-array key arrays keep it fast for realistic frames.

## Summary by CodeRabbit

* **New Features**
  * Multi-key sorting UI: toolbar sort-strip, header/context-menu controls, keyboard shortcuts, visual priority badges, and sort-status messaging.
  * Optional per-panel sort persistence (raven.dataViewer.persistSort).

* **Behavior**
  * Stable, full-width integer and mixed-type ordering; NA/NaN always placed last; labels/format toggle affects sort keys.
  * Row responses include original row indices when sorted.

* **Documentation**
  * Expanded Data Viewer sorting docs, settings reference entry, and comparison table.

* **Tests & Fixtures**
  * Comprehensive sort/persistence tests and new arrow fixtures covering bigints/labelled columns.

[Review Change Stack](https://app.coderabbit.ai/change-stack/jbearak/raven/pull/326?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


docs/data-viewer.md
```diff
@@ -156,19 +158,83 @@ same `View(mtcars)` opened tomorrow remembers the layout.
...
+when picking *Sort ascending* / *Sort descending* is a shortcut for the
+*Add* items.
+
+A sorted column shows a hairline triangle on the right edge of its
+header — ▲ for ascending, ▼ for descending. When more than one column
+is in the sort, each sorted header also shows a small priority badge
+(1, 2, 3 …) so you can see which key takes precedence at a glance.
+
+A chip strip appears in the toolbar listing the active keys in
+priority order; each chip has a kebab (▾) that opens a small popover
```
jbearak: I removed the kebab and made the pill clickable

docs/data-viewer.md
```diff
@@ -156,19 +158,83 @@ same `View(mtcars)` opened tomorrow remembers the layout.
...
+  label exists for a cell) when Labels is on.
+- **Numeric** columns always sort by the underlying double, even when
+  the Format toggle is rounding the display.
+
+Toggling Labels with a sort active on a labelled column will re-sort
+the rows.
+
+### Persistence
+
+Sort state is persisted per panel-name + schema-hash alongside layout
```
jbearak, line 216: I don't think it's safe to assume that schema+row count = same values. We need to recompute the row order

jbearak:
Filed the gutter / per-row marker work as #327 (enhancement). The wire payload is already in place — only the webview render path needs the synthetic-column refactor described there.

PR #354, opened 2026-05-31
docs: CI package-exports database (planned) — user + architecture docs

### LINKED ISSUE #350: Resolve package datasets (lazy-data) as symbols, not just base datasets
## Problem

Bare references to a package's **datasets** are flagged as undefined variables — for example `library(nycflights13); flights` or `ggplot(diamonds, aes(...))`. This happens **even when R is present**, because dataset names are not part of a package's `getNamespaceExports()` and Raven currently does not surface them for non-base packages:

- `get_package` sets `lazy_data = Vec::new()` for regular packages (`crates/raven/src/package_library.rs:1236`).
- Resolution (`is_symbol_from_loaded_packages`, `collect_exports_recursive` at `:1468`) checks a package's `exports`, never `lazy_data`.
- Only **base** packages get their datasets resolved, because `initialize()` (`:1021`) merges `parse_data_symbols()` output into `base_exports` for the base set. That's why `mtcars` / `iris` work but `flights` / `diamonds` do not.

So package-dataset references are a standing false-positive source in the undefined-variable diagnostic.

## Goal

Make a package's datasets resolvable as symbols for **installed packages**, consistently in the editor (LSP) and the CLI (`raven check`), with no regression to function/export resolution.

## Approach (sketch)

1. **Populate `lazy_data` in `get_package`.** For a regular package whose directory is found, call `parse_data_symbols(pkg_dir)` (`crates/raven/src/namespace_parser.rs:510`) and store the result in `PackageInfo.lazy_data` instead of `Vec::new()` (`package_library.rs:1236`). Mirror the symlink-safety handling already used for base packages in `initialize()` (`:1017-1038`).
2. **Fold `lazy_data` into the resolvable set.** `collect_exports_recursive` (`:1468`) should add `package_info.lazy_data` alongside `exports` (`:1499`) so `get_all_exports` / `combined_exports` include datasets; the sync checks (`is_symbol_from_loaded_packages`, `find_package_for_symbol`) then resolve dataset names through the existing cache path.
3. **Keep the model honest.** Keep datasets in `lazy_data` (distinct from `exports`) — don't merge into `exports`; just include both in resolution. This preserves the functions-vs-data distinction for hover/completion attribution.

## Performance

`parse_data_symbols` walks `data/` (filesystem I/O), but `get_package` is cached, so the cost is one-time per package on the async resolution path — acceptable, and off the synchronous diagnostic hot path.

## Testing

- A fixture package exposing a dataset resolves the dataset name (no undefined-variable).
- Datasets resolve transitively through `Depends` and meta-package `attached_packages` (same recursion as exports).
- No regression: existing function/export resolution tests stay green.

## Out of scope

Dataset *columns* / structure, and signatures. This is dataset **object names** only.

## Relationship to the CI package-exports database

This is a **prerequisite** for the CI package-exports DB work (`docs/superpowers/specs/2026-05-30-ci-package-exports-db-design.md` and its plan). That design captures `lazy_data` in the Tier 2/3 records but deliberately does **not** wire dataset *resolution*, to avoid an editor-vs-CI asymmetry (CI resolving datasets that a local R install would not). Once this issue lands, the DB's captured `lazy_data` is consumed by the same resolution path automatically, so CI resolves package datasets via the providers with no extra work. Implement this issue first, then the CI DB plan.

jbearak:
## Stage 1 of the CI package-exports database — documentation-first

This is the **docs-only Stage 1** of the CI package-exports DB feature, shipped as its own PR *before any feature code* (per the plan's "Execution order — documentation-first"). It is the **agreed UX/behavior contract**: reviewers approve the intended command names, flags, file locations, the names-vs-install-status split, and the three tiers here, and the implementation (Stages 2–6) builds to match in later sessions.

**No feature code** is touched — only `docs/` and the `CLAUDE.md` doc-map (its real file, `AGENTS.md`, which `CLAUDE.md` symlinks to). Every new/changed user-facing doc/section carries a `> **Status: planned.**` banner; the final code milestone (Task 6.5) removes those banners once the feature ships.

### Authoritative artifacts
- **Plan:** `docs/superpowers/plans/2026-05-30-ci-package-exports-db.md` (Milestone 6, Tasks 6.0–6.4 + 6.4b)
- **Design spec:** `docs/superpowers/specs/2026-05-30-ci-package-exports-db-design.md`
- **Prerequisite (landed):** raven#350 — package-dataset (lazy-data) resolution

### What's in this PR (Tasks 6.0–6.4)
- `docs/development.md` — architecture/internals: the `package_db` module, the `PackageMetadataProvider` seam (Tier 1 bespoke + first; Tier 2/3 on a `get_package` miss; `package_exists()` stays Tier-1-only; why `prefetch_packages` is unchanged), performance discipline (index preload, lazy mmap decode, `blake3` verify in `spawn_blocking`), the Tier 3 build pipeline (reference-R full library ∪ CRAN+Bioc r-universe, append-only, `names-db` GitHub Release, base-exports companion), the #350 dependency, and the testing approach.
- `docs/cli.md` — `raven packages freeze` (`--used` default / `--installed`/`--all`, `--output`, `--workspace`; renv-first order; renv.lock as a set selector; no-op when unchanged), the maintainer-only `raven packages build-shipped-db`, and `raven check --report-uninstalled` (missing-package off by default in CI; the flag reports `library()` calls absent from the local lib paths — *not* relative to Tier 2/3).
- `docs/diagnostics.md` — names ≠ install status (§10): missing-package is Tier-1-only and independent of the export DB; per-mode behavior table; the accepted typo gap.
- `docs/cross-file.md` — the three-tier export resolution; Tiers 2/3 are export-names-only (no `:::`, no signatures).
- `docs/r-package-dev.md` — generating `.raven/packages.json` for CI and the renv-first library order.
- `docs/package-database.md` (new) — the full user page: three tiers + fidelity table, the committed Tier 2 file (generated/reviewable, no-op-when-unchanged, version-skew explained), the maximally-inclusive `--used` set, the bundled `names.db` (append-only reference-R ∪ CRAN+Bioc, GitHub Release, `blake3`, weekly+release cadence), the base-exports file, package datasets via #350, and the fidelity caveats. Linked from the `CLAUDE.md` doc-map.

### Verification
- `git diff --name-only main...HEAD` → only `AGENTS.md` + `docs/*` (no `crates/`, no `editors/vscode/src`, no workflows, no `package.json` config).
- No new LSP setting → settings-reference generator deliberately **not** run.
- Status banner present in every new/changed doc; all cross-reference anchors verified against existing headings.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -5,6 +5,7 @@ Raven ships a single binary that serves the LSP via stdio *and* exposes subcomma
 - `raven check` — index a workspace and report the **full** diagnostic set (the same diagnostics the editor publishes), for CI gating.
 - `raven lint` — run the native **style** linter only.
 - `raven analysis-stats <path> [--csv] [--only <phase>]` — profile workspace analysis phases (`scan`, `parse`, `metadata`, `scope`, `packages`); see `raven --help`.
+- `raven packages freeze` — generate a committed `.raven/packages.json` export database so `raven check` can resolve package symbols in CI when no R or packages are installed (see [`raven packages`](#raven-packages) and [Package database](package-database.md)). The maintainer-only `raven packages build-shipped-db` builds Raven's bundled `names.db`.
 
 The difference between `check` and `lint` is **scope**: `lint` parses each file in isolation and runs only the style rules, so it needs no R installation and no workspace index — but it can't see relationships between files. `check` builds the same workspace index the language server builds (and, unless packages are disabled, runs R to resolve installed-package exports and base R symbols), so it additionally reports cross-file, undefined-variable, and package diagnostics. `lint` is therefore the cheaper, R-free option for pure style gating; `check` does more work per run in exchange for the editor's full analysis in CI.
 
@@ -21,7 +22,7 @@ Diagnostics reported (subject to configured severities — see [diagnostics.md](
 - Syntax errors and semantic checks (e.g. assignment-in-condition, mixed logical operators).
 - The native style lints (when enabled via `raven.toml` / `.lintr`).
 - Cross-file diagnostics: missing sourced files, circular dependencies, exceeded max source-chain depth, redundant directives, and out-of-scope usage.
-- Missing-package warnings (`library(notInstalled)`).
+- Missing-package warnings (`library(notInstalled)`) — see [Missing-package reporting in CI](#missing-package-reporting-in-ci) for the planned change to suppress these by default.
 - Undefined-variable diagnostics, accounting for cross-file and package scope.
 
 ### Workspace and paths
@@ -38,6 +39,7 @@ The whole workspace is always indexed so cross-file resolution is accurate. The 
 - `--no-config` — ignore `raven.toml` and `.lintr`; use Raven's built-in defaults.
 - `--format text|json|sarif` — default `text`.
 - `--max-severity off|hint|info|warning|error` — highest severity that does **not** fail the build (default `info`). With the built-in defaults, undefined-variable and missing-file diagnostics are `warning` and circular dependencies are `error`, so they fail the build at the default threshold.
+- `--report-uninstalled` — re-enable missing-package warnings, which `raven check` otherwise suppresses by default. See [Missing-package reporting in CI](#missing-package-reporting-in-ci).
 - `--quiet` — suppress the trailing summary line.
 - `--color auto|always|never` — when to colorize `text` output (default `auto`). See [Color output](#color-output).
 - `--no-color` — alias for `--color never`.
@@ -47,6 +49,16 @@ The whole workspace is always indexed so cross-file resolution is accurate. The 
 `raven check` auto-detects R on `PATH` to resolve installed-package exports and base R symbols (it runs `.libPaths()` and parses package `NAMESPACE` files, the same as the language server). It honors the same `raven.toml` package settings the editor does: `packages.enabled = false` disables R detection entirely (no R subprocess, no package or base-symbol diagnostics — matching the editor), `packages.rPath` selects the R binary instead of `PATH` auto-detection, and `packages.additionalLibraryPaths` adds extra library directories to the search path. If R is not found, package and base-symbol diagnostics are limited — `library()` calls aren't checked against installed packages, and undefined-variable detection falls back to a built-in symbol list — and a one-line note is printed to stderr. All other diagnostics still run.
 
 Before reporting, `raven check` warms the export cache for the packages each reported file attaches with `library()` / `require()`, so a bare call into an attached package that isn't one of its exports is flagged the same way the editor flags it. One narrow gap remains: a package attached only *indirectly* — in a `source()`d file rather than in the reported file itself — is not pre-warmed, so calls that could resolve to such a package are left unflagged rather than risk a false positive. Attach the package directly in the file (or rely on the editor) if you need those calls checked.
+
+### Missing-package reporting in CI
+
+> **Status: planned.** Describes the CI package-exports database, in active development; not yet in a released build. Tracking: the package-database work (and prerequisite [raven#350](https://github.com/jbearak/raven/issues/350)).
+
+`raven check` resolves package **export names** through an ordered three-tier fallback — installed packages, then a committed `.raven/packages.json`, then Raven's bundled `names.db` — so symbols from attached packages resolve even when no R is installed. This stops the undefined-variable storm that otherwise makes Raven unusable in CI. See [Package database](package-database.md).
+
+Knowing a package's exports is **separate** from knowing whether it is installed. The **missing-package** diagnostic answers a different question — *"will `library(X)` succeed at runtime?"*, i.e. is `X` installed? — so it stays **Tier-1-only**: it is driven solely by what is present in the local library paths, never by the export database. Because CI deliberately omits package installation, `raven check` **suppresses missing-package warnings by default**.
+
+`--report-uninstalled` re-enables them. It is useful when a pipeline *does* install packages (e.g. `renv::restore()`) and wants to fail if any didn't install. It reports `library()` calls **not present in the local library paths** — **not** relative to the Tier 2/Tier 3 export metadata. One consequence: with the default off, a genuine typo such as `library(dpylr)` is silent (no tier knows it, but `raven check` isn't checking install status); it is reported only with `--report-uninstalled`. The language server is unchanged — it still fires missing-package whenever install state is known. See [Diagnostics](diagnostics.md#package-names-vs-install-status) for the full model.
 
 ### File encoding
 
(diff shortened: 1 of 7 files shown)


docs/development.md
```diff
@@ -274,6 +274,161 @@ All R subprocess calls must:
...
++ Bioc r-universe (`cran.r-universe.dev` and `bioc.r-universe.dev` per-package
+`_exports`/`_dependencies`/`_datasets`). Precedence: **reference-R > r-universe >
+retained-from-prior**; a rebuild never drops a package (decision #8). The full
+installed library is the point — hard-to-install / GitHub-only / internal
+packages enter the floor only this way, and append-only keeps them. The
+maintainer **bootstraps** the floor once from a richly-provisioned machine; teams
+can build a private `names.db` from their own library for richer in-house
+coverage.
+
+- **Entry point:** `raven packages build-shipped-db` (maintainer/CI-only — the
```
jbearak, line 390: `build-shipped-db` should append by default to a previously built ship db.

docs/r-package-dev.md
```diff
@@ -249,6 +249,22 @@ imports remain visible if some `R/*.R` files don't carry roxygen tags.
 
 Raven watches for changes to `DESCRIPTION` and `NAMESPACE` files. After running `devtools::document()` or editing these files directly, diagnostics update automatically without restarting the editor.
 
+## Generating a package database for CI
+
+> **Status: planned.** Describes the CI package-exports database, in active development; not yet in a released build. Tracking: the package-database work (and prerequisite [raven#350](https://github.com/jbearak/raven/issues/350)).
+
+To gate diagnostics on your package in CI without installing R packages there, generate a committed export database and check it in:
```
jbearak, line 256: This sentence isn't right. You get diagnostics no matter what. The reason to run `raven packages freeze` is:

1. in case u use any packages whose exports aren't shipped with `raven`
2. in case any of those packages' exports differ from the versions used to build raven's shipped exports database, in ways that could affect your diagnostics

docs/package-database.md
```diff
@@ -0,0 +1,101 @@
...
+
+## The three tiers
+
+| Tier | When it applies | Source | Fidelity |
+|---|---|---|---|
+| **1 — Installed** | Package installed locally (and R reachable for `exportPattern`) | The existing on-disk path: static `NAMESPACE` parse + an R subprocess to expand `exportPattern` | Authoritative, version-exact to the install |
+| **2 — Repo DB** | Not installed, or R can't launch | A repo-specific, committed `.raven/packages.json` you generate locally | "Frozen Tier 1": full structure, version-exact to when it was generated |
+| **3 — Shipped DB** | Otherwise | A Raven-bundled `names.db` built from a reference-R capture ∪ latest CRAN + Bioconductor | Latest CRAN/Bioc; export names only |
+
+**Tier 2 outranks Tier 3** because it is project-specific and built through the authoritative path. A repo that never generates a Tier 2 file still works in CI via Tier 3 alone (latest-only). The two databases share one in-memory model, one reader, and one writer.
```
jbearak, line 17: Drop " (latest-only)". I don't think it's clear what that means -- that Tier 3 has the latest version of the package, as of raven's release. What I think we might want to say is that if u use a newer version of the package than is in Raven's Tier 3 database, then, it's possible you'll get a false positive diagnostic, if you use a symbol that wasn't in the version that was published when Raven was last shipped

docs/package-database.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Tier 3 — bundled `names.db`
+
+A Raven-bundled, latest-CRAN/Bioc export database — the R-free floor that works with no setup at all.
+
+- **Source.** An authoritative **reference-R capture of the build machine's entire installed library** (via Raven's Tier-1 path) **∪** **CRAN + Bioconductor** from [r-universe](https://r-universe.dev) (`cran.r-universe.dev` and `bioc.r-universe.dev`). r-universe's `_exports` is the *true post-load namespace export set* — `exportPattern()` already expanded — for the whole ecosystem. The two sources are merged **append-only**: each rebuild starts from the previous database, overlays the reference-R capture, appends r-universe packages not already present, and **never drops** a package. Precedence: reference-R > r-universe > retained-from-prior.
```
jbearak, line 64: What does "Precedence" mean here? I think this points to a flaw in our plan. I think that if a newer version of a package is in CRAN/bioc, then, we should use that version's exports. (As written, this text suggests we'd use the older set of symbols from reference-R even if a newer set of symbols was available)

docs/diagnostics.md
```diff
@@ -59,6 +59,28 @@ If the symbol is defined later in the same file at top level, the message also r
...
+
+- **Export resolution** (suppresses undefined-variable noise) uses all three tiers, in every mode.
+- **Install status** (drives the *missing-package* diagnostic) is **Tier 1 only** — it reflects what is actually present in the local library paths, and never the export database. A database that knows `dplyr`'s exports does **not** make `dplyr` count as installed.
+
+#### Per-mode behavior
+
+| | Export resolution | Missing-package ("not installed") |
+|---|---|---|
+| **Language server (interactive)** | tiers 1→2→3 (no undefined-variable storm even when R is absent) | Fires when install state is known and the package is absent — regardless of the database. Unchanged from today; the database stops the symbol storm but never masks the "install this dependency" nudge. |
+| **`raven check` (CI)** | tiers 1→2→3 | **Suppressed by default** (CI deliberately omits installation). Re-enable with [`--report-uninstalled`](cli.md#missing-package-reporting-in-ci). |
```
jbearak, line 76: Let's not have `raven check` suppress diagnostics. Let's just have `raven ci` be an alias to `raven check --suppress-package-warnings` (not sure what to call this --suppress option) so that a user can still run raven check locally to get useful info

jbearak (commented):
i debated reporting uninstalled packages in `raven check`, but, afaict, `check` is the convention used in CI, and unlike for software dev, i don't think we expect a social science CI workflow to actually run R scripts. OTOH, this -would- be relevant for R package development.

I think we should document that, if your workflow is actually to run your R scripts in CI, in which case `raven check` would be followed by actually running your scripts, then, you actually *do* want to see uninstalled packages warnings.

I'm really unsure what the default should be. I think reporting them by default sort of goes against the point of having all this infra to be aware of package exports. Therefore, though I suggested, in my inline comments, adding a `raven ci` command, which by default suppressed package installation warnings, and having `raven check`, by default, emit warnings for uninstalled packages, I think, upon reflection, that we should stick with the original plan, and have `raven check` suppress these by default.

docs/cli.md
```diff
@@ -5,6 +5,7 @@ Raven ships a single binary that serves the LSP via stdio *and* exposes subcomma
 - `raven check` — index a workspace and report the **full** diagnostic set (the same diagnostics the editor publishes), for CI gating.
 - `raven lint` — run the native **style** linter only.
 - `raven analysis-stats <path> [--csv] [--only <phase>]` — profile workspace analysis phases (`scan`, `parse`, `metadata`, `scope`, `packages`); see `raven --help`.
+- `raven packages freeze` — generate a committed `.raven/packages.json` export database so `raven check` can resolve package symbols in CI when no R or packages are installed (see [`raven packages`](#raven-packages) and [Package database](package-database.md)). The maintainer-only `raven packages build-shipped-db` builds Raven's bundled `names.db`.
```
jbearak, line 8: there's supposed to be an option for `freeze` to tell it whether to resolve all packages on one's system, or only those used by the repo

jbearak:
Closing unmerged. #354 was selected as the base for the Stage 1 docs; its commits have been consolidated onto the working branch `claude/epic-borg-3a6061` (merged there, not into `main`), alongside the post-review spec/plan revisions. The docs will get a drift-correction pass to match the revised design (version-aware Tier 3 merge, uncommitted refresh interval, concrete drift wording, corrected `freeze` rationale).

PR #366, opened 2026-06-02
names.db: dated reproducible releases; unbundle from the binary

jbearak:
## Summary

Two related changes to how Raven's `names.db` package-export sidecar is released and consumed, plus a docs/README pass.

### Dated, reproducible `names.db` releases
- `build-names-db.yml` now publishes an **immutable `names-db-YYYY-MM-DD`** release on each build (retained indefinitely) alongside the existing rolling `names-db` tag.
- `raven packages update [YYYY-MM-DD]` gains an optional dated positional: with no argument it pulls the latest (`names-db`); with a date it pulls that immutable snapshot for reproducibility. The date is validated and mutually exclusive with `--base-url`.

### `names.db` no longer ships with the binary
- `release-build.yml` no longer bundles `names.db` into the release archives (they now contain only the `raven` binary + `LICENSE`). The VSIX already excluded it.
- Added a `names-db-compat` preflight job: it validates the about-to-ship `raven` against the **current** `names-db` Release via `raven packages validate-shipped-db`. This is a compatibility gate (does the shipping Raven read the live DB?), not a bundling step; `build-lsp` depends on it.

### Docs / README
- De-jargoned `docs/cli.md` (bare "Tier N", "zero-adoption", "Base-7", set-builder `∪` notation), removed the fictional "package-manager builds ship names.db" claim across all live docs, moved maintainer commands + `--base-url` self-hosting into a cli.md appendix, and added a **"Why Raven analyzes R without running it"** section.
- Flipped every "bundled alongside the binary" claim to "download via `raven packages update`" across `cli.md`, `README.md`, `package-database.md`, `development.md`, `r-package-dev.md`, `cross-file.md`, `diagnostics.md`.
- Documented the one-level-deep `$` accessor-completion limit in `limitations.md`.
- Rewrote the README intro into a tighter pitch (scope model, CI gap, NSE).

## Testing
- `cargo test -p raven --lib parse_update_args` — 5 passed (new dated-arg parser tests).
- `cargo build -p raven` — clean.
- Both `release-build.yml` and `build-names-db.yml` validated as YAML.

## Notes
- The `names-db-compat` gate downloads the `names-db` Release asset, so a raven release now depends on that asset existing (it's already bootstrapped). Can be softened to skip-with-warning on a missing asset if desired.

## Summary by CodeRabbit

* **New Features**
  * `raven packages update` now accepts an optional YYYY-MM-DD date argument to access specific database snapshots
  * Introduced immutable, date-stamped snapshot releases of the names database

* **Documentation**
  * Enhanced README with clearer messaging about static scope resolution and Rust performance
  * Expanded CLI documentation with detailed CI workflows and package metadata supply guidance
  * Clarified names database distribution model across documentation
  * Added limitation documentation for nested member completions


README.md
```diff
@@ -1,12 +1,10 @@
...
-
-Raven's static analysis also reaches into accessors: start typing `fruit$a` and it can suggest `apple` the moment you open the file, with no R session — Positron offers the same only from a live session (Raven's accessor analysis currently goes one level deep). The [comparison page](docs/comparison.md) lays out the differences in detail, including where the existing tools cover ground Raven doesn't.
+Doing this well in R is hard — part of why so little static tooling exists for the language. R leans heavily on non-standard evaluation (NSE): a function can take its arguments as unevaluated code and decide what the symbols mean at runtime, so `col` in `dplyr::filter(df, col > 0)` is a data-frame column, not an undefined variable. A checker that didn't account for this would flag idiomatic R as errors; Raven recognizes these NSE contexts and leaves them unflagged. The same analysis isn't only defensive: wherever it can determine an object's structure statically, Raven completes its fields — start typing `fruit$a` and it can suggest `apple` the moment you open the file, with no R session.
```
jbearak: We should couch this. It detects NSE contexts, but, there's code outside NSE context that depends on NSE, and Raven will report false positives, if, e.g., you use features like `attach()`. If you use that, then,  you need to declare the variables you attached. We should state this concisely ,and provide a link to the directives doc

jbearak: My basic view on this is I don't, and have never, in well over a decade of using R, relied on such features as would cause a false positive. TypeScript vs JavaScript is the best analogy I can think of, but that analogy goes too far so it might not be the right example to give.

docs/development.md
```diff
@@ -328,10 +328,9 @@ the seam.
   raises `window/showMessage`, then resolution degrades to the next tier. A
   missing or unreadable database never hard-fails the LSP or the build.
 - **Tier 3 locator order:** environment overrides still win, then the user-data
-  sidecars installed by `raven packages update`, then exe-relative sidecars shipped
-  by release archives and package-manager builds. Source/Cargo
-  installs normally have only the user-data candidate unless the sidecars were
-  placed manually next to the executable.
+  sidecar installed by `raven packages update`, then an exe-relative sidecar
+  (e.g. one placed next to the binary by hand). Installs normally have only the
+  user-data candidate, since `names.db` is no longer bundled next to the binary.
```
jbearak: s/no longer bundled next to/not bundled with/

docs/development.md
```diff
@@ -435,16 +434,16 @@ coverage.
   `workflow_dispatch` + scheduled job
   (`.github/workflows/build-names-db.yml`; weekly schedule `0 6 * * 1`) downloads
   the current asset (the seed), rebuilds via the shared
-  `scripts/build-names-db.sh`, and re-uploads. `release-build.yml` downloads the
-  same asset to place `names.db` exe-relative next to the binary in the release
-  archives; the VSIX omits it (VS Code users resolve their locally installed
-  packages via Tier 1). A Git LFS seed is committed at
+  `scripts/build-names-db.sh`, and re-uploads. `release-build.yml` no longer
+  bundles `names.db`: release archives ship only the `raven` binary, and the VSIX
+  omits it too (VS Code users resolve their locally installed packages via Tier 1).
```
jbearak, line 439: get rid of the "no longer" since we never shipped a release that did this

README.md
```diff
@@ -1,12 +1,10 @@
...
 
-The language server analyzes your code in realtime: it completes variable and accessor names as you type, flags syntax errors and undefined variables — including use-before-definition, because scope is position-aware even within a single script (a variable defined on line 50 isn't in scope on line 10) — and lets you jump to where a variable or function is defined or list all the other places that your codebase references it.
+Raven is fast — written in Rust — and runs all of this in realtime in your editor, with diagnostics updating instantly as you type. It also fills a gap at the other end of the workflow: automated checks in pull requests — running `raven check` flags undefined variables and other errors before code merges, the kind of automated review that's table stakes for other languages but has long been missing for R.
```
jbearak: s/Raven is fast — written in Rust —/Raven is fast/

("Rust" is an implementation detail, and you can have fast w/o Rust, and not all readers will no what Rust is, since R is used by people who aren't software engineers. Also, Raven is written in Rust + TypeScript, although the LSP itself if written in Rust. Another reason not to say this is Raven's sister project Sight, my Stata LSP, is written solely in TypeScript is also fast--so Rust isn't the only route to speed.)

jbearak: Done — the line now reads "Raven is fast and runs all of this in realtime in your editor…", dropping the "— written in Rust —" aside. (2d24e15d)

docs/development.md
jbearak: Done — dropped "no longer"; it now reads "`release-build.yml` does not bundle `names.db`: release archives ship only the `raven` binary…". (2d24e15d)

README.md
jbearak: Couched it. The sentence now reads: "…Raven recognizes these NSE contexts and leaves them unflagged — though code that pulls names into scope at runtime, such as `attach()`, can still draw a false positive, which you resolve by declaring those symbols with a [directive](docs/directives.md)." I left out the TypeScript/JavaScript analogy per your follow-up. (2d24e15d)

docs/development.md
jbearak: Done — now reads "since `names.db` is not bundled with the binary." (2d24e15d)

Issue #404, opened 2026-06-06
foreach


jbearak:
`foreach (i = 1:10) { print(i) }` loops have undefined variable false positives because Raven doesn't understand that in `foreach` loops, between the parentheses, you defined an iterator `i` that's in scope between the foreach's `{}`'

this is a regression that must've been caused by #401

PR #409, opened 2026-06-06
Model nested foreach composition with %:% and when() filters (#406)

### LINKED ISSUE #406: Model nested foreach composition with %:% and when() filters
Follow-up to #404.

#404 covers the minimal regression fix: named non-dot iterator arguments from `foreach(...)` are visible inside the RHS of `%do%` / `%dopar%`.

Nested foreach composition needs a separate design because `%:%` and `when()` introduce additional binding/filter scopes before the final execution operator:

```r
foreach(i = 1:3) %:% foreach(j = 1:3) %do% i + j
foreach(i = 1:3) %:% when(i %% 2 == 0) %do% i
```

Follow-up acceptance criteria:

- how iterator bindings flow across `%:%`;
- whether `when(...)` filter expressions see iterators from the left side;
- how multiple nested foreach levels compose for `%do%` and `%dopar%`;
- leak behavior after the composed expression.

jbearak:
Closes #406. Follow-up to #404.

## What

Extends `foreach` iterator scoping to nested compositions joined by the `%:%` operator and `when(...)` filters. The left side of `%do%`/`%dopar%` may now be a `%:%` chain of `foreach(...)` calls and `when(...)` filters:

```r
foreach(i = 1:3) %:% foreach(j = 1:3) %do% i + j
foreach(i = 1:3) %:% when(i %% 2 == 0) %do% i
foreach(i = 1:3) %:% when(i %% 2 == 0) %:% foreach(j = 1:3) %do% i + j
```

## Design decisions (the issue's acceptance criteria)

- **Iterator bindings across `%:%`** — every `foreach()` call in the chain contributes its named, non-dot iterators; all are visible in the executed body.
- **`when(...)` filters see left-side iterators** — the synthetic iterator scope spans the whole left-hand composition (new `ForeachExecution::scope_start`), so iterators resolve inside `when(...)` filter expressions.
- **Multiple nested levels** — composition recognition is recursive and associativity-agnostic; `%do%` and `%dopar%` behave identically.
- **Leak behavior** — iterators do not leak past the composed expression; body-local definitions remain scope-local, exactly as in #404.
- **`when(...)`** is a filter, not an iterator source, so it contributes no iterators.

A simple `foreach(...) %do% body` keeps #404 behavior verbatim (scope starts at the rhs, not the foreach call).

### Deliberate tradeoff

The composition scope spans the iterator-value expressions so the common cross-reference idiom (`foreach(j = seq_len(i))` referencing the outer `i`) is not false-flagged. The accepted cost: a symbol *assigned* inside a left iterator-value expression (the pathological `foreach(i = { x <- 1; 1:3 }) %:% ...`) is treated as loop-scoped and does not leak. This is documented in `ForeachExecution::scope_start` and `docs/diagnostics.md`.

## Implementation

- `crates/raven/src/foreach.rs` — recursive `collect_composition_iterators` over the `%:%` chain; `is_colon_chain` predicate (single source of truth for `%:%` nodes); `lhs_is_foreach_call` generalized to `is_foreach_package_call(node, text, name)` covering both `foreach` and `when` (bare + `foreach::`/`foreach:::`).
- `crates/raven/src/cross_file/scope.rs` — scope event uses `exec.scope_start` for the interval start.
- Recognition stays syntax-only (no package metadata), so editor and CLI behave identically.

## Tests

Coverage at all three layers (recognition, scope resolution, diagnostics): nested composition, `when()`-filter visibility, multi-level chains, `%dopar%`, cross-iterator references, no-leak, body-local non-leak, and the iterator-value-expression check boundary. Rejection cases for `when()`-only and non-foreach operands.

## Gates

`cargo fmt --all --check`, `cargo clippy --workspace --all-targets --features test-support -- -D warnings`, and the full `cargo test -p raven --lib` suite (4421 tests) all pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved R foreach() iterator scoping for nested compositions and filters; iterator visibility, binding, shadowing, and left-to-right evaluation behave correctly.

* **Tests**
  * Added regression tests covering nested foreach/%:% compositions, when() filters, scoping and undefined-name detection.
  * Updated editor toolbar layout test to validate wrap/badge boundary and layout behavior.

* **Documentation**
  * Updated diagnostics to reflect new foreach scoping rules.


jbearak:
Follow-up commit (41e7704a) addresses the review finding that the composition scope was too wide and lost left-to-right binding order.

**Before:** one `FunctionScope` spanned the whole `%:%` composition with all iterators as flat parameters. Since function params are position-independent within a scope, every iterator was visible everywhere in the chain. The diagnostics path already respected definition order (so `raven check` was correct), but the shared scope read by hover / go-to-definition / completion over-exposed: a reverse reference `foreach(i = seq_len(j)) %:% foreach(j = 1:3)` showed `j` as visible in the first foreach's value, and in `foreach(i=1:3) %:% when(i>1) %:% foreach(i=4:6)` the `when()`'s `i` resolved to the inner declaration instead of the left one.

**After:** one `FunctionScope` per `foreach(...)` call, each spanning *[end of that call → shared body end]*. In a `%:%` chain these nest (a later call's scope is contained in earlier ones), reproducing foreach's left-to-right binding via ordinary closure + shadowing semantics:

- an iterator is visible in later `when()` filters, later iterator value expressions (`foreach(j = seq_len(i))` still resolves the outer `i`), and the body — but not to its left, so the reverse reference is correctly flagged, matching R's "object not found";
- the body resolves a shadowed name to the innermost (rightmost) `foreach`; a `when()` filter resolves it to the nearest `foreach` on its left.

New scope-level tests cover the reverse-reference exclusion and both shadowing cases. All gates pass (`cargo fmt --all --check`, clippy `-D warnings`, full `cargo test -p raven --lib` — 4424 tests).

PR #419, opened 2026-06-07
Fix package cache contention

### LINKED ISSUE #414: Sync cache readers treat try_read() contention as a cache miss (transient wrong results)
## Summary

The synchronous, cached-only readers on `PackageLibrary` use `try_read()` and treat a *contended* lock the same as a genuine cache miss. Under lock contention (a concurrent write while a read is attempted) they can return a transient wrong answer rather than the correct cached one:

- `is_symbol_from_loaded_packages` → returns `false` ("not a package export"), which can surface a **transient false-positive "undefined variable" diagnostic**.
- `find_package_owner_for_symbol` / `get_owned_exports_for_completions` → return `None` / drop symbols, briefly losing **hover/help/signature attribution and completions** for attached symbols.

Affected methods (all in `crates/raven/src/package_library.rs`): `is_symbol_from_loaded_packages`, `get_exports_for_completions`, `get_owned_exports_for_completions`, `find_package_for_symbol`, `find_package_owner_for_symbol`.

These methods are deliberately synchronous (diagnostic collection can't `.await`) and use `try_read()` to avoid blocking the runtime — the fail-soft is intentional. The defect is only that a *contended* lock is indistinguishable from *"symbol genuinely absent,"* so the conservative fallback can produce a user-visible false positive instead of deferring.

## Severity: low

- **Pre-existing.** The `try_read`-fail-soft pattern predates the #407 `combined_entries` unification. #407 merged two read locks into one, so a single write now blocks both projections at once — a slightly wider window, but that is the intended atomicity tradeoff, not a regression.
- **The worst window is already mitigated.** Diagnostic collectors gate on `snapshot.package_library_ready` (`handlers.rs:5126`), and the codebase explicitly prefers a transient false positive over over-suppression for the narrow out-of-scope collector (`handlers.rs:5118-5125`). So the startup window is largely covered; the residual exposure is contention *after* warm-up (e.g. an invalidation burst from a libpath change while actively editing).
- **Self-healing.** Diagnostics are recomputed and republished on the next revalidation, so any false flag clears on its own.

## How hard to address

**Not hard to mitigate well; the root-cause fix is a medium refactor.** Options, cheapest first:

1. **Document only (trivial).** Note the post-warm-up contention window as a known limitation. Closest to status quo.
2. **Three-state return so contention defers instead of false-flagging (medium, localized).** Change `is_symbol_from_loaded_packages` (and peers, as needed) to distinguish *"lock contended / unknown"* from *"definitely not a package export"* (e.g. return an enum, or `Option<bool>`), and have the diagnostic collectors treat *unknown* as "don't flag" — consistent with the existing `package_library_ready` philosophy. Touches the return type plus ~2 diagnostic call sites in `handlers.rs`. This kills the false-positive symptom without changing the locking model.
3. **Lock-free snapshot reads (medium-large, root cause).** Replace the `tokio::RwLock<HashMap<…>>` caches with `arc_swap::ArcSwap<Arc<HashMap<…>>>` (or hand-rolled `ArcSwap`). Readers do a wait-free atomic load — they never block and never see a contention "miss"; writers clone-and-swap. This eliminates the issue for *all* sync readers, removes #407's contention-widening, and is a small read-path perf win. Cost: change the field types and every accessor, adjust the async write paths (`get_all_exports`, `invalidate_many`, `clear_cache`) to swap instead of `write().await`, and add the `arc-swap` dependency.

Approaches that DON'T work here:
- `blocking_read()` panics when called from within the async runtime (these readers run inside async handlers) — not viable.
- Switching to `std::sync::RwLock` risks a guard held across `.await` in the async write paths — viable but ripply and easy to get wrong.

**Recommendation:** if addressed, option 2 (defer-on-contention) is the cheap, targeted symptom fix; option 3 (ArcSwap) is the principled long-term fix and pairs naturally with #413 (which also wants `get_all_exports` to hand back a shared snapshot rather than a clone). Filed as `defer` to decide later.

## Context

Surfaced during the multi-angle review of the #407 `combined_entries` unification. Related: #413 (avoid cloning the aggregate export set; an `ArcSwap`/snapshot model would address both).

jbearak:
## Summary
- Fixes #414 by switching PackageLibrary package caches to `parking_lot::RwLock` and using blocking cached reads so lock contention is not treated as package/export absence.
- Keeps cache lock scopes short by cloning relevant `Arc` handles under guards and doing completion materialization outside the locks.
- Adds targeted bounded diagnostic fan-out for dependency-triggered republishing while preserving global ordered LSP dispatch.
- Documents the new cache-lock and diagnostic fan-out invariants in `docs/development.md`.

## Validation
- `cargo test -p raven bounded_fanout::run_bounded_fanout_runs_all_items_without_exceeding_limit`
- `cargo test -p raven sync_reader`
- `cargo fmt --all --check`
- `cargo test -p raven`

## Artifacts
- Warp conversation: https://app.warp.dev/conversation/78ffda55-d951-406d-866d-de7b413ea111
- Plan: https://app.warp.dev/drive/notebook/c1MQ5S2ttP0JPgJSU2ZIYF

## Summary by CodeRabbit

* **Refactor**
  * Implemented bounded-concurrency diagnostic publishing across document and file-watch events to prevent resource exhaustion and improve stability under high parallelism
  * Reworked package-cache into snapshot-based reads with serialized publication to reduce contention and make lookups more deterministic

* **Chores**
  * Updated developer documentation to reflect concurrency handling and cache locking guidance

* **Tests**
  * Added tests validating bounded fan-out behavior and cache snapshot correctness

* **Benchmarks**
  * Added benchmarks for package-cache lookups and diagnostic fan-out performance

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -106,6 +106,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a23eb6b1614318a8071c9b2521f36b424b2c83db5eb3a0fead4a6c0809af6e61"
 
 [[package]]
+name = "arc-swap"
+version = "1.9.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6a3a1fd6f75306b68087b831f025c712524bcb19aad54e557b1129cfa0a2b207"
+dependencies = [
+ "rustversion",
+]
+
+[[package]]
 name = "arrayref"
 version = "0.3.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1604,6 +1613,7 @@ version = "0.9.0"
 dependencies = [
  "ammonia",
  "anyhow",
+ "arc-swap",
  "async-trait",
  "blake3",
  "bson",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1617,6 +1617,7 @@ dependencies = [
  "lru",
  "memmap2",
  "notify",
+ "parking_lot",
  "percent-encoding",
  "postcard",
  "proptest",
--- a/crates/raven/Cargo.toml
+++ b/crates/raven/Cargo.toml
@@ -21,6 +21,7 @@ name = "raven"
 path = "src/main.rs"
 
 [dependencies]
+arc-swap = "1"
 ammonia = "4"
 blake3 = "1"
 bson = "2"
--- a/crates/raven/Cargo.toml
+++ b/crates/raven/Cargo.toml
@@ -91,6 +91,10 @@ required-features = ["test-support"]
 name = "edit_to_publish"
 harness = false
 required-features = ["test-support"]
+[[bench]]
+name = "package_cache_and_fanout"
+harness = false
+required-features = ["test-support"]
 
 [[example]]
 name = "profile_diagnostics"
--- a/crates/raven/Cargo.toml
+++ b/crates/raven/Cargo.toml
@@ -48,6 +48,7 @@ ropey = "1.6"
 streaming-iterator = "0.1"
 lru = "0.16"
 notify = { version = "6.1", default-features = false, features = ["macos_fsevent"] }
+parking_lot = "0.12"
 percent-encoding = "2"
 tempfile = "3"
 toml = "0.8"
--- a/crates/raven/benches/package_cache_and_fanout.rs
+++ b/crates/raven/benches/package_cache_and_fanout.rs
@@ -13,7 +13,7 @@ use raven::package_library::{PackageInfo, PackageLibrary};
 use std::collections::HashSet;
 use std::sync::{Arc, mpsc};
 use std::thread;
-use std::time::Duration;
+use std::time::{Duration, Instant};
 use tokio::runtime::{Builder, Runtime};
 
 const PACKAGE_COUNT: usize = 24;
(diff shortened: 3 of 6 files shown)


jbearak:
## Targeted benchmark summary

I added and ran a targeted Criterion suite for the issue #414 paths rather than relying on the startup benchmark, which mostly exercises unrelated parsing/scan work. The suite covers package-cache lookups, warmed aggregate lookups, completion materialization, publication/contention behavior, and bounded diagnostic fan-out.

### Key findings

- The original bug was correctness-sensitive: old synchronous package-cache readers could treat cache contention as absence and return false/empty results. The snapshot refactor removes that failure mode by reading immutable `ArcSwap` snapshots.
- The final snapshot implementation makes the publication/contention benchmark effectively disappear from the hot path:
  - `package_cache/publication_gate/time_to_correct_symbol_read`: **128.1 µs → 910 ns** (**99.3% faster**) in the 1,000-sample comparison against old `main` with the same benchmark harness.
- Completion materialization improved materially because the current implementation clones relevant cached handles and preallocates the result map before expanding strings:
  - `package_cache/uncontended/completions_24x400_exports`: **929.2 µs → 679.6 µs** (**26.9% faster**).
- Ordinary cached package lookup paths are roughly neutral to slightly faster:
  - `uncontended/is_symbol_hit_24_loaded`: **541.4 ns → 530.3 ns** (**2.1% faster**)
  - `uncontended/find_owner_hit_24_loaded`: **556.3 ns → 549.6 ns** (**1.2% faster**)
  - `uncontended_combined/is_symbol_hit_24_loaded`: **545.4 ns → 518.8 ns** (**4.9% faster**)
- The warmed aggregate owner lookup is slightly slower but still sub-microsecond:
  - `uncontended_combined/find_owner_hit_24_loaded`: **767.6 ns → 792.9 ns** (**3.3% slower**)
- Diagnostic fan-out microbenchmarks are unchanged by the package-cache snapshot work, as expected:
  - serial fan-out: **+0.1%**
  - bounded fan-out: **+0.1%**

### Validation

After the snapshot refactor, these passed locally:

- `cargo test -p raven sync_reader`
- `cargo bench -p raven --features test-support --bench package_cache_and_fanout --no-run`
- `cargo test -p raven bounded_fanout::run_bounded_fanout_runs_all_items_without_exceeding_limit`
- `cargo fmt --all --check`
- `cargo test -p raven`

Overall: the targeted benchmarks show the refactor fixes the contention-as-cache-miss class of bug and substantially improves the affected publication/contention path, while keeping steady-state lookup performance broadly neutral and improving completion-heavy reads.

PR #424, opened 2026-06-10
fix: make system.file source resolution recoverable across package lifecycle events

jbearak:
## Problem

`source(system.file("x.R", package = "pkg"))` resolution had an unrecoverable-staleness design gap. `resolve_system_file_sources` collapsed resolution state: a successful hit cleared the `SystemFileCall` marker, and a failed attempt (package not installed, `lib_paths` non-empty) **dropped the source entry from metadata entirely**. Since `resolve_system_file_in_workspace` only revisited entries with `system_file.is_some()`, the collapse was permanent:

1. **Install after startup** — a `LibpathEvent::Changed` for a newly-installed package invalidated caches and republished, but a previously-dropped system.file edge never formed until the user edited the sourcing file or restarted. Symmetrically, package **removal** left a stale `resolved_uri` and dependency edge.
2. **`Package:` rename** — the DESCRIPTION manifest branch updated the workspace package name but never re-resolved, so branch-1 self-package references could not recover.

## Fix

Retain, don't collapse — resolution state is recomputed idempotently on every pass:

- `system_file` is **never cleared** and unresolved entries are **never dropped**. A branch-2 miss with `lib_paths` available clears any stale `resolved_uri`/`path` but keeps the entry, inert (no edge, no missing-file diagnostic), for later re-resolution. The four resolution states are documented on `resolve_system_file_sources`.
- `resolve_system_file_in_workspace` returns the URIs whose resolution actually changed and rebuilds graph edges only for those. `system_file_republish_set` expands them with **open transitive dependents** (same `compute_affected_dependents_after_edit` fan-out as `did_change`) — a parent's scope traverses child edges, so a child's new edge changes the parent's diagnostics.
- The `LibpathEvent::Changed` consumer re-resolves behind a cheap `WorkspaceIndex::any_entry` read-lock pre-check (workspaces without matching system.file sources never take the write lock), using a **package-filtered** variant so only entries referencing changed packages are cloned or disk-probed, then force-republishes affected open docs (monotonicity-safe force-republish path).
- The DESCRIPTION/NAMESPACE manifest branch re-resolves after `apply_package_event`, extending the republish set through the existing gated `extend_affected_for_manifest_change` helper.
- The missing-file-diagnostics skip is now a single predicate, `ForwardSource::exempt_from_missing_file_diagnostics`, replacing a condition previously duplicated across both collectors. Branch-1 `/inst/...` paths stay diagnosable.
- Change detection resolves into a cloned sources `Vec` (new slice-based `resolve_system_file_source_entries`), paying for `Arc::make_mut` and re-insertion only on a real change; `WorkspaceIndex::entries_matching` snapshots only the matching subset.

## Tests

New `lifecycle` module in `crates/raven/tests/system_file_source.rs` (all written red-first):
- edge forms after package install without editing the file
- stale edge/`resolved_uri` clear after removal; recovery on reinstall
- branch-1 recovery after a `Package:` rename
- package-filtered resolution leaves unrelated resolved entries untouched
- open dependents of changed files are in the republish set

Plus unit tests for the diagnostics-exemption matrix and the new `WorkspaceIndex` filters. Existing unit tests that encoded the old clear/drop semantics were updated deliberately.

## Gates

- `cargo fmt --all --check` clean
- `cargo clippy --workspace --all-targets --features test-support -- -D warnings` zero warnings
- `cargo test -p raven --features test-support`: 4719 lib tests + all integration suites, 0 failures

Docs updated: `docs/development.md` (retention/lifecycle/republish mechanics) and `docs/cross-file.md` (live tracking of installs/removals/renames).

## Summary by CodeRabbit

* **New Features**
  * `system.file()` paths now automatically re-resolve when packages are installed, removed, or renamed during a session, eliminating the need for restarts.

* **Bug Fixes**
  * Improved handling of unresolved paths that can now recover when referenced packages become available.
  * Enhanced accuracy of missing-file diagnostics for system.file() sources.

* **Documentation**
  * Updated documentation describing system.file() resolution behavior across package lifecycle events.


jbearak:
Addressed the outside-of-diff CodeRabbit finding in 3257fad9 — the concern was valid.

**Verified:** `get_enriched_metadata` reads document-store metadata in preference to the workspace index, so re-resolving only index entries left open buffers stale. The graph-edge-first lookup in scope resolution masked it when buffer text matched the on-disk index entry, but it broke for edited-unsaved buffers and entirely when the file isn't indexed (unsaved buffer, `index_workspace = false`).

**Fix:** `resolve_system_file_in_workspace_for_packages` now runs a second pass over open-document metadata (same clone-resolve-compare flow; new `DocumentStore::replace_metadata` swaps the metadata Arc in place — tree/artifacts don't derive from source path resolution, so no re-parse). The pass runs after the index pass so buffer-derived graph edges win for files present in both stores, and changed open buffers are reported in the returned republish list. `index_cross_package_resolved_files` and the libpath consumer's `any_entry` pre-check consider open buffers as well. Since every event arm goes through this one shared function, both flagged sites (manifest branch and `LibpathEvent::Changed`) are covered.

**Test:** `lifecycle::open_buffer_metadata_re_resolves_after_package_install` (red-first) — an open buffer absent from the workspace index re-resolves after a package install event, reports its URI for republish, and gains a dependency edge from buffer metadata.

jbearak:
Follow-up in a24f5dfc: for a file present in both stores, the index pass could rebuild the dependency graph from stale index metadata, and the open-document pass's early `continue` (buffer resolution unchanged) skipped the authoritative rebuild — e.g. a buffer that resolved at did_open while the scanned index entry was still unresolved kept index-derived edges until the next edit.

The open pass now tracks which URIs the index pass rebuilt and re-runs `update_file` with the buffer's metadata for unchanged buffers among them, so buffer-derived edges always win. Unchanged-and-untouched files still skip all graph work. Pinned by `lifecycle::open_buffer_graph_edges_win_over_index_rebuild_when_unchanged` (red-first: the stale edge at the index call site survived; now the buffer call site wins).

PR #440, opened 2026-06-12
Enumerate lazy-loaded package data via R subprocess (issue #429)

### LINKED ISSUE #429: Enumerate lazy-loaded package data via R subprocess (Rdata.rdb datasets, data() aliases; sysdata.rda package-mode only)
## Context

Corpus triage (#423, PR #427) left ~73 ledger entries that all reduce to one capability gap: Raven cannot enumerate objects stored in R's serialized lazy-data databases, even for **installed** packages.

- 44 — `lazy-loaded dataset of a library()-attached package (bundled in Rdata.rdb, not enumerable from data/ files)`: e.g. broom tests do `library(survival); lung` — survival is installed, but its `data/` ships only `Rdata.rdb`/`Rdata.rdx` with no `datalist`, so dataset names are invisible (probe-verified during triage).
- 26 — `internal data from R/sysdata.rda` (cli's `emojis`, googledrive's `.endpoints`, readr's `date_symbols`).
- 3 — `object loaded by data(name, package=) call where the data file defines differently-named objects` (survey's `data/api.rda` defines `apiclus1`/`apistrat`; filename-stem discovery yields only `api`).

PR #427 added an embedded-table floor for the base `datasets` package (INDEX topics under-enumerate multi-object entries), but that only covers shipped base packages.

This overlaps with the recommended-group `data()`/`library()` remaining-work item — landing this first would shrink that triage substantially.

## Scope decisions

**1. Installed-package exported datasets (Rdata.rdb) — in scope.** `library(survival); lung` is legitimate R usage even though `lung` is only discoverable through `Rdata.rdb`/`Rdata.rdx` or `data(package = "survival")`. This is the 44-entry bulk.

Semantic detail to get right: the presence of `Rdata.rdb` in an installed package's `data/` directory is itself the LazyData marker — R builds that lazyload DB only when `DESCRIPTION` sets `LazyData`, and only then does `library(pkg)` actually attach the datasets. For non-LazyData packages (raw `.rda` files in `data/`), `library(pkg); dataset` is an error in real R; those names bind only through explicit `data()` calls.

**2. Package-internal `R/sysdata.rda` — in scope only for package-development mode** (checking that package's own code). sysdata objects live in the namespace unexported: `library(cli); emojis` is a genuine error in real R (even `cli::emojis` fails; access requires `:::`). They must never be injected into user-script scope after `library(pkg)`.

Raven already has package-mode sysdata machinery (`crates/raven/src/package_state/sysdata.rs`: AST-first extraction of `use_data(..., internal = TRUE)` / `save(..., file = ".../sysdata.rda")` calls, R-subprocess `load()`+`ls()` fallback, digest-keyed caching). First step for this bucket: **re-probe the 26 entries** — they may predate that machinery or stem from runs without R available. Any entry where a *user script* references sysdata objects after `library()` should be **reclassified as a true positive**, not cleared.

**3. Multi-object data-file aliases — in scope.** `data(api, package = "survey")` creating `apiclus1`/`apistrat` is unusual but valid; filename-stem discovery cannot see it. The right ceiling is R's own behavior: R can enumerate multi-object files only when the package ships a `datalist` or a LazyData index, so matching R's `data(package=)` listing is the bar — not exceeding it.

## Spec

**Approach: use R as the oracle** (option 2 of the original three). Enumerate via the existing R subprocess, cache the result, degrade gracefully without R — the same hybrid shape already established for `exportPattern` expansion (static parse first; batched subprocess query where statics can't express the answer — see `get_multiple_package_exports` in `crates/raven/src/r_subprocess.rs`; fallback when R is absent). Since the goal is matching R semantics, asking R is cleaner than reimplementing its lazyload internals.

### Enumeration

Batched R query for installed packages that need it (those whose `data/` exists): `data(package = "pkg")$results[, "Item"]`. The `Item` column encodes aliases as `"apiclus1 (api)"`, so one query covers buckets 1 and 3 together. Parse into per-package `{object_name, file_stem}` pairs.

- Follow the `r_subprocess` module-doc safety invariants: validate package names, wrap every call in a timeout, never interpolate strings into R code.
- Cache at the `PackageLibrary` level (existing `ArcSwap<PackageCache>`), alongside the other per-package metadata.
- Without R: fall back to today's behavior (file-stem discovery + the embedded base floor from #427). No regression, reduced fidelity.

### Consumption — two distinct points, do not conflate

1. **Post-`library()` scope injection.** Inject enumerated object names into scope after `library(pkg)`/`require(pkg)` for LazyData packages (`Rdata.rdb` present). Feed `PackageInfo.lazy_data` so completion attribution stays distinct from exports. Whether to additionally *tighten* the existing static file-stem path for non-LazyData packages (where `library()` alone does not attach datasets in real R) can be decided during implementation — it risks flipping currently-suppressed entries into true positives and needs a corpus re-probe first.
2. **`data()` alias expansion.** Extend the existing recognizer (`try_extract_data_call_definitions`, `crates/raven/src/cross_file/scope.rs`), which today binds only the literal argument names: `data(api)` / `data(api, package = "survey")` should bind every enumerated object whose file stem is `api`, falling back to the literal name when enumeration is unavailable. Bucket 3 lands entirely here.

### Out of scope / later

- **Option 3 is mostly moot: `names.db` already carries dataset lists.** `PackageRecord.lazy_data` is part of the Tier 2/3 schema, and the r-universe `_datasets` ingest fills it with authoritative names, including `Rdata.rdb`-bundled ones (e.g. dplyr → `starwars`/`storms`). The genuinely open follow-ups are consumption-side:
  - Tier 2/3 providers are consulted only on a Tier-1 miss (no installed package directory — see `get_package`'s fallback in `crates/raven/src/package_library.rs`), so the DB's dataset lists never help the *installed*-package case behind the 44 entries. Consulting DB `lazy_data` as a no-R floor for installed packages (with the version-skew caveat) is the real offline follow-up.
  - The `build-shipped-db --capture-reference` source routes through `get_package` and so today shares Tier 1's static file-stem blindness; on version ties its records beat r-universe's in the append-only merge, which can shadow authoritative dataset lists with incomplete ones. Landing subprocess enumeration in `get_package` upgrades the reference capture — and removes that merge wrinkle — for free.
- **Option 1** (parse `Rdata.rdx`/RDS in Rust) — last resort. It couples Raven to R's serialization/lazyload internals, and that format family has a history of sharp edges (e.g. the CVE-2024-27322 promise-injection class). Revisit only if the no-R gap proves to matter in practice.

## Acceptance criteria

- `library(survival); lung` produces no undefined-variable diagnostic when survival is installed (editor + `raven check`).
- `data(api, package = "survey")` followed by `apiclus1` produces no undefined-variable diagnostic when survey is installed.
- Negative: `library(cli); emojis` in a plain user script still flags — sysdata must not leak into user-script scope.
- sysdata bucket: the 26 entries re-probed; entries covered by package-mode machinery cleared, user-script-level references reclassified as true positives.
- Cleared ledger entries pruned; full 4-group strict run stays green. (The corpus environment has R installed — "survival is installed" implies R present — so subprocess-based enumeration suffices for the gate; option 1 is not required for CI.)

## Pre-PR review protocol (multi-subagent, required)

When implementing this issue, do **not** open a PR until the branch survives **two consecutive clean review passes**. Orchestrate each pass the way `/code-review` works — parallel dimension reviewers plus adversarial verification:

**Prerequisites for any pass:** implementation complete; `cargo fmt --all --check`, `cargo clippy --workspace --all-targets --features test-support -- -D warnings`, and `cargo test -p raven` all green. For corpus-affecting changes, the relevant strict run (`RAVEN_CORPUS_GROUPS=...`) must report 0 unclassified / 0 stale acceptances / 0 stale FPs.

**One review pass:**
1. Dispatch parallel reviewer subagents over the full branch diff (`git diff <base>...HEAD`), one per dimension:
   - **Correctness** — logic bugs, edge cases, AST/scope handling errors;
   - **Invariants** — CLAUDE.md "Key invariants" plus the module/function doc-comment invariants of every touched file;
   - **Tests** — red→green discipline, per-surface coverage (editor diagnostics, `raven check`, Rmd chunks where applicable), negative cases that keep genuine diagnostics firing;
   - **Docs** — user-facing `docs/` updates and doc comments per CLAUDE.md documentation requirements;
   - **Simplification/reuse** — duplicated logic, missed existing helpers.
2. Each reviewer reports findings with `file:line`, severity, and confidence.
3. **Adversarially verify** every finding with an independent subagent instructed to refute it against the actual code; discard refuted findings.
4. The pass is **clean** iff zero findings survive verification.

**The gate:** fix all surviving findings, re-run the gates, then run a **full fresh pass** — any fix resets the counter. Open the PR only after two clean passes in a row (the second confirms the fixes introduced nothing new).

jbearak:
Closes #429.

## What this does

Raven can now enumerate objects stored in R's serialized lazy-data databases for installed packages, closing the largest remaining corpus-triage gap (~710 ledger entries):

- **Enumeration** — new batched R-subprocess query `get_multiple_package_datasets` (`data(package = "pkg")$results[, "Item"]`), parsing `name (stem)` alias entries, following the same hybrid shape as `exportPattern` expansion. Falls back to today's static file-stem discovery (+ INDEX + embedded base floor) without R.
- **Population** — enumerated object names replace `PackageInfo.lazy_data` for LazyData packages (`data/Rdata.rdb` presence is the marker; only then does `library(pkg)` attach datasets in real R). A new `PackageInfo.data_aliases` map (file stem → bound object names) is filled for every data-shipping package. `library(survival); lung` resolves via the existing `collect_exports_recursive` folding.
- **`data()` alias expansion** — new `ScopeEvent::DataLoad` records `data()`/`utils::data()` calls (stems + string-literal `package=`); query-time expansion binds the enumerated objects across the recursive and streaming resolvers: `data(api, package = "survey")` binds `apiclus1`/`apistrat`; bare `data(api)` searches packages attached at-or-before the call plus base. Literal names always remain bound (no-R fallback). The `package=` argument feeds the cross-file interface hash. `raven check` warms packages named only in `data(package=)`.
- **Own-package sysdata** — a package's own code can reference its `R/sysdata.rda` objects in `raven check`; sysdata never leaks into user scripts (`library(cli); emojis` still flags — pinned by an R-gated negative test).
- **R-spawn concurrency cap** — a global semaphore bounds concurrent R subprocesses (permit acquired outside the per-call timeout). Net win: the full lib test suite runs in ~25s vs ~109s on main, with the R-timeout flakes gone.
- **Corpus** — 710 stale false-positive ledger entries pruned (677 lazy-data/alias + 33 sysdata); full 4-group strict run green (0 unclassified / 0 stale acceptances / 0 stale FPs). Checkpoint doc updated.

## Acceptance criteria (issue #429)

- `library(survival); lung` — no undefined-variable diagnostic (editor + `raven check`) ✅
- `data(api, package = "survey")` then `apiclus1` — clean ✅
- Negative: `library(cli); emojis` in a user script still flags ✅
- sysdata entries re-probed and resolved; ledger pruned; strict corpus green ✅

## Review protocol

Per the issue's pre-PR requirement: parallel dimension reviews (correctness, invariants, tests, docs, simplification) with adversarial verification of every finding, repeated until **two consecutive fully-clean passes**. Earlier passes surfaced and fixed: an enumeration-site duplication, missing `extract_data_packages`/CLI-warming/late-arm/clamp tests, silent acceptance-test skips, a doc renumbering, and the test-suite contention regression that led to the semaphore.

Docs: `docs/cross-file.md`, `docs/package-database.md`, `docs/limitations.md`, `docs/development.md`, `docs/package-corpus-checkpoint.md`.

Out of scope (per design): names.db `lazy_data` as a no-R floor for installed packages; Rust-side `Rdata.rdx` parsing; tightening non-LazyData `library()` injection (separate follow-up; needs corpus re-probe).

## Summary by CodeRabbit

* **New Features**
  * Expand package `data(...)` aliases so dataset names resolve from packages when available; batched enumeration populates those aliases.
  * Fail-soft fallback to discover package-local sysdata when initial scan finds none.

* **Bug Fixes**
  * Fewer false positives for dataset and sysdata-based undefined-symbol diagnostics.

* **Performance**
  * Global throttling for R subprocesses to avoid oversubscription.

* **Documentation**
  * Expanded docs on lazy-loaded datasets, sysdata, and developer guidance.


jbearak:
Follow-ups from the review of this PR (not blockers; the branch's reviewed fixes are committed):
- #441 — pre-existing cleanups surfaced during review.
- #442 — `data()` alias follow-ups: local-def precedence discrepancy, function-scoped exclusion coverage, recursion robustness.
- #443 — investigate the parse-under-parallel-load nondeterminism that blocked landing some of the #442 tests.

PR #471, opened 2026-06-15
fix(cross-file): isolate each parent frame's forward sources in prefix walk

jbearak:
## Problem

Raven emitted false `undefined-variable` diagnostics for a symbol defined in an **earlier-sourced sibling**. Reported on the `worldwide` repo: `data.r` sources `covariates.r` (which defines `ww.covars`) before sourcing `indices.r`, yet `ww.covars` was flagged undefined throughout `indices.r`.

## Root cause

In `parent_prefix_at`, each parent frame was resolved with `isolate_forward_source_visits = false`, so the frame's **backward (parent) walk** and its own **forward `source()` resolution** shared a single `visited` map.

When `indices.r`'s parent prefix queried `data.r`, that frame's backward walk — reaching `data.r` again at the EOF sentinel through a hub file (`functions.r`) shared by the grandparent and another script — expanded the earlier-sourced sibling `covariates.r` into `visited` at `(MAX, MAX)`. `data.r`'s own STEP-2 source of `covariates.r` then hit the revisit guard and returned an empty scope, dropping `ww.covars` from the prefix.

The trigger requires a dense graph (a shared hub creating a deep parent walk), which is why it reproduced on the real repo but not on a handful of files.

## Fix

Recurse into parent frames with `isolate_forward_source_visits = true`, so each parent frame's forward sources resolve against a pre-backward-walk `visited` snapshot — exactly like the queried file's own forward sources already do. The backward walk still shares the live map for cycle pruning; only STEP-2 forward children are isolated.

One-line behavioral change plus doc-comment updates at both the call site and the `forward_visited_base` snapshot site.

## Test

`sibling_source_symbol_visible_through_shared_hub_parent_walk` — a minimal 6-file reproduction (the hub structure). Fails without the fix (`got symbols: ["helper"]` — hub symbol survives, sibling's `ww.covars` dropped), passes with it.

## Verification

- Full lib suite: **5132 passed, 0 failed**
- `cargo fmt --all --check` clean; `cargo clippy --workspace --all-targets --features test-support -- -D warnings` clean
- `worldwide` repo: `indices.r` false positives **8 → 0**; whole-workspace diagnostic count otherwise unchanged (the fix only ever resolves more symbols, never adds diagnostics)

## Performance

The fix trades aggressive (incorrect) `visited` pruning for correct re-resolution of forward sources. Measured on `worldwide` (a dense, hub-heavy graph; release build):

| Path | Unfixed | Fixed | Δ |
|---|---|---|---|
| `raven check` whole workspace | ~5.01s | ~5.35s | **+7%** |
| `raven check` single deep file | ~0.68s | ~0.68s | **~0% (noise)** |

The latency-sensitive per-file editor path is unaffected (the queried file's prefix is cached and cross-file scope is a small slice of per-file work). The ~7% lands on the full-workspace CI sweep, bounded by `max_chain_depth` and `max_transitive_dependents_visited`. A lazy-clone experiment confirmed the cost is the re-resolution, not the map clones. Follow-up to eliminate it (memoize forward-source child contributions) tracked separately.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where symbols defined in earlier-sourced siblings were not properly visible across certain code hierarchy patterns.

* **Tests**
  * Added a regression test to verify correct symbol visibility through parent walk scenarios.

### DIFF
--- a/crates/raven/src/cross_file/scope.rs
+++ b/crates/raven/src/cross_file/scope.rs
@@ -4789,7 +4789,19 @@ where
             hoist_globals,
             backward_dep_mode,
             is_cancelled,
-            false,
+            // Isolate this parent frame's own forward source() visits from its
+            // own backward (parent) walk. A parent like `data.R` resolved here
+            // runs both STEP 1 (its grandparent walk) and STEP 2 (its own
+            // `source()` children). Without isolation, the grandparent walk can
+            // expand one of `data.R`'s children — a sibling sourced earlier,
+            // e.g. `covariates.R` — into the shared `visited` map at the EOF
+            // sentinel, after which `data.R`'s own STEP 2 source of that child
+            // hits the revisit guard and returns an empty scope, silently
+            // dropping its definitions from the prefix. Passing `true` makes
+            // STEP 2 here use a pre-STEP-1 `visited` snapshot, exactly as the
+            // queried file's own forward sources do. Regression:
+            // `sibling_source_symbol_visible_through_shared_hub_parent_walk`.
+            true,
             None,
             None,
             // Parent-prefix walk does not expand `data()` aliases (issue #429):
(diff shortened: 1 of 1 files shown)


jbearak:
Performance follow-up (memoize forward-source child contributions to recover the ~7% whole-workspace regression): #472

PR #518, opened 2026-06-21
fix(data-viewer): compute filter histograms lazily (fixes ~1min empty-grid on large frames)

jbearak:
## Problem

`View()` on a large frame (the `demo/data-viewer-smoke.R` 10M-row × 50-col case) showed an empty grid — "Showing 0-0 of 0" — for **~1 minute** before the data appeared. This regressed a previously-instant, virtualized load.

## Root cause

It was a **performance** regression, not a correctness bug. The grid's row fetching is still virtualized (`ArrowSliceReader.open()` returns in ~0.3s with the right `nrow`/columns). But `DataViewerPanel.sendInit`/`sendReplace` precomputed a 50-bin histogram for **every** numeric column *before* posting the `init` message that lets the grid paint. Each column costs two full scans, so on the 10M×50 frame this blocked first paint for **~49s** behind a full-frame read.

Measured on a real 3.73 GB Arrow file:

| Init phase | Time |
|---|---|
| `ArrowSliceReader.open()` | 266 ms |
| `computeNumericHistograms()` (all columns) | **48,853 ms** |

The eager scan also re-decoded all batches ~100× due to the 16-entry LRU thrashing under the column-outer/batch-inner loop.

## Fix: lazy per-column histograms

Histograms only feed the filter popover's range brush, so they don't need to exist before the grid paints — or at all for columns the user never filters.

- **init/replace ship no histograms**; the grid paints immediately.
- The webview requests a column's histogram (`getHistogram`) the first time its filter popover opens; the host computes that one column on demand (`computeHistogramForColumn`, ~0.66s for a 10M-row column), caches it per-reader, and replies (`histogram`). Mirrors the existing `getLabels`/`labels` on-demand round-trip.
- Caches cleared on dataset change (host on `replace`; webview on `replace`/non-same-dataset init).

| | Before | After |
|---|---|---|
| First paint | ~49 s (blocked) | **~0.3 s** |
| Open a numeric filter | (already loaded) | ~0.66 s, then cached |
| Non-numeric columns | scanned anyway | 0 |

## Why it regressed silently — test added

The previous suite asserted histograms were **present** in `init`, which locked in the slow path. That's replaced with a regression guard asserting **`sendInit` triggers zero batch decodes** (it fails the instant any full-frame scan creeps back onto the paint path), plus round-trip tests for the on-demand fetch and its degradation paths (decode-throw, out-of-range index, non-numeric column).

## Hardening (from review)

The `getHistogram` handler is a webview trust boundary: it now always posts a reply (out-of-range / non-numeric / decode error all degrade to empty bins — never a stranded in-flight marker or unhandled rejection), follows the `getRows` disposal discipline, and gates scans on a single shared `isNumericArrowType` classifier (the webview fetch gates on `colKind`, so "shows a brush" and "fetches bins" can't diverge).

## Verification

- `bun run typecheck` clean; 379 data-viewer bun tests pass; bundle builds; bundle-size test passes.
- Two consecutive clean `/code-review` passes on the final commit; codex adversarial review (findings addressed).
- Docs: `docs/data-viewer.md` updated for the on-demand histogram behavior.

## Release Notes

* **Performance & Optimization**
  * Improved initial table load times by computing numeric column histograms on-demand (when filter editors are opened) rather than upfront. Histograms are cached after first computation for consistent performance.

* **Documentation**
  * Updated data viewer documentation to clarify lazy histogram computation behavior.

### DIFF
--- a/docs/data-viewer.md
+++ b/docs/data-viewer.md
@@ -279,7 +279,11 @@ and applying updates the existing filter (it does not add a second one — see
 
 For numeric `between` / `not between`, the editor shows a histogram of the
 column's distribution with draggable range thumbs alongside the numeric
-input fields.
+input fields. The histogram is computed the first time you open that
+column's filter (not up front when the table loads), so on a very large
+frame it may appear a moment after the popover opens; it is cached
+thereafter. This keeps the grid itself painting immediately regardless of
+row count.
 
 For factor columns, the editor shows a searchable checklist of levels when
 the column's value dictionary has been shipped; large or unshipped
--- a/docs/data-viewer.md
+++ b/docs/data-viewer.md
@@ -282,8 +282,9 @@ column's distribution with draggable range thumbs alongside the numeric
 input fields. The histogram is computed the first time you open that
 column's filter (not up front when the table loads), so on a very large
 frame it may appear a moment after the popover opens; it is cached
-thereafter. This keeps the grid itself painting immediately regardless of
-row count.
+thereafter. This keeps the histogram scan off the initial load path, so
+opening a table paints the grid without waiting for every numeric column to
+be scanned.
 
 For factor columns, the editor shows a searchable checklist of levels when
 the column's value dictionary has been shipped; large or unshipped
--- a/editors/vscode/src/data-viewer/histograms.ts
+++ b/editors/vscode/src/data-viewer/histograms.ts
@@ -28,16 +28,23 @@ import type { HistogramBin } from './messages';
 
 const BIN_COUNT = 50;
 
+/** True for the Arrow types that produce a numeric histogram (the same
+ *  types `colKind` classifies as `numeric` / `labelledNumeric`). The host
+ *  uses this as a trust-boundary guard before launching an on-demand scan
+ *  for a webview-supplied column index. */
+export function isNumericArrowType(arrowType: string): boolean {
+    return arrowType.startsWith('Int')
+        || arrowType.startsWith('Uint')
+        || arrowType.startsWith('Float');
+}
+
 export async function computeNumericHistograms(
     reader: ArrowSliceReader,
 ): Promise<Record<number, HistogramBin[]>> {
     const out: Record<number, HistogramBin[]> = {};
     const schema = reader.schema.columns;
     for (let ci = 0; ci < schema.length; ci++) {
-        const t = schema[ci].arrowType;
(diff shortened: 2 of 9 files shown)


jbearak:
Added a follow-up commit: the toolbar now shows **"Loading…"** instead of "Showing 0-0 of 0" until the `init`/`replace` message lands (and suppresses the matching "0 rows × 0 columns" status-bar text). Before `init`, `nrow` is 0 and the bare "0-0 of 0" reads as an empty/broken table — most noticeable on a large frame where `open()`/first paint takes a beat. Implemented via an optional `loading` flag on the pure `describeVisibleRows` helper (unit-tested) plus a `loading` state in `App.tsx` cleared on init/replace. Verified: 380 data-viewer bun tests pass, typecheck clean, bundle builds, one clean review pass.

jbearak:
### Added: remove the redundant bottom status bar (`887c5e57`)

Follow-on cleanup, stacked on the loading-indicator commit (it removes the same `statusText` block that one touched, so it lives here rather than in a separate PR off `main`).

**Why:** the bottom status bar was introduced incidentally by the React port (`0f927311`) — the original Svelte viewer had only a toolbar counter. Its contents now duplicate existing toolbar UI:

| Status-bar piece | Toolbar equivalent |
|---|---|
| `N rows x M columns` total | row-count lead `Showing X-Y of N` |
| `K columns hidden` | `Columns ⓚ` badge |
| `sorted by …` | sort chip strip |
| `filtered to N` | filter chips + filtered total in the lead |
| object class (`data.frame`) | — (dropped; not wanted) |

**What changed:**
- Removed the status bar and its `statusText`/`sortStatusText`/`filterStatusText` and `.status-bar` CSS.
- Relocated the **only** non-duplicated piece — the transient `Sorting…`/`Filtering…` progress cue — to a pill in the toolbar chip group, driven by a new pure helper `pendingOperationLabels()` (unit-tested). It renders independently of the chip strips, so it shows even on the first sort/filter (verified in review). Pulse animation honors `prefers-reduced-motion`.
- Removed the object-class report end-to-end (R bootstrap → arrow-reader → protocol → panel → webview) and deleted the orphaned `shape-description.ts` (+ test).

**Verification:** TDD for the helper; full suite green (1331 bun + 317 vscode-integration), typecheck + bundle clean. Two code-review passes (first found 2 low-severity stale-comment/CSS leftovers, fixed; second clean) + a clean codex adversarial pass.

Issue #562, opened 2026-07-04
Raw URI identity lets symlink/case aliases bypass open-document-authoritative checks


jbearak:
Raven treats LSP document identity, watcher events, dependency-graph nodes, indexes, caches, and diagnostics gates as raw `Url` keys. That is consistent internally for normal same-URI document lifecycles, but it means an open buffer under one spelling of a file path may not protect the same underlying file when a watcher event or `source()` edge arrives under another spelling — a symlink target or a different case.

Concretely (from a Codex-assisted audit during #561 review):

- Open docs, `DocumentStore`, both workspace indexes, the file cache, `DependencyGraph` nodes, and the diagnostics gate are all keyed by raw `Url` (`state.rs:496`, `workspace_index.rs:126`, `cross_file/file_cache.rs:73`, `cross_file/dependency.rs:557`, `cross_file/revalidation.rs:85`).
- The #535/#476 case-leniency is path-resolution machinery only: it corrects resolved-path casing below trusted prefixes so edge targets match index keys, and deliberately avoids `std::fs::canonicalize` because resolving symlinks would diverge from uncanonicalized index keys (`cross_file/path_resolve.rs:662,685`). `path_to_uri` is plain `Url::from_file_path` (`path_resolve.rs:936`).
- Source-edge resolution derives a child URI from the parent URI's path context with lexical `.`/`..` normalization and no symlink canonicalization (`path_resolve.rs:61,738`, `dependency.rs:724`) — so a `source()` edge through `/link/...` produces graph target `/link/helper.R` while the same file open under `/real/helper.R` is a distinct open-doc key, and "open docs are authoritative" checks miss.
- Whether VS Code's watcher itself can emit alternate spellings of a `didOpen`'d URI could not be confirmed from this repo; the extension and `vscode-languageclient` forward watcher URIs without normalization, and `backend.rs:2177` already acknowledges noncanonical watcher paths for symlinked roots in package routing.

This predates #558/#561 — the close-resync path added there is same-URI safe; it is simply another consumer of the existing convention.

Decision to make: add a canonical/alias identity layer for open-document-authoritative checks (watched-files skip, resync veto, `update_from_disk` guard), or explicitly document raw-URI identity as the supported LSP model and add targeted tests pinning symlink/case alias behavior.

Origin: flagged as a hypothesis in review of #561; assessed as pre-existing/systemic rather than a PR gap.

https://claude.ai/code/session_01GkFAeWVzq3XaA2dKy2jze7

PR #568, opened 2026-07-05
feat(config): project-level exclusions for indexing and raven check

### LINKED ISSUE #559: Add project-level exclusions for workspace indexing and raven check
## Background

Sight added config-driven directory/glob exclusions in https://github.com/jbearak/sight/pull/256.

Raven currently has lint-only `.lintr` exclusions, but workspace indexing and `raven check` still mostly rely on hardcoded skip directories.

Relevant Raven paths:

- `crates/raven/src/state.rs` `scan_workspace` / `collect_files_matching`
- `crates/raven/src/cli/check.rs`
- `crates/raven/src/config_file/`
- `docs/configuration.md`
- `docs/cli.md`

## Problem

Large generated, vendored, output, or archived directories can still be indexed and checked unless they happen to match Raven hardcoded skip list. That can add noise and cost, especially in projects with generated R files or large external trees.

This is distinct from lint overrides: users may want Raven to ignore a tree for indexing, dependency discovery, diagnostics, and default CLI discovery, not just disable lint rules inside it.

## Desired behavior

Add a project-level exclusion setting, likely in `raven.toml`, with project-root-relative glob semantics.

Suggested behavior:

- apply exclusions to workspace scans and default `raven check` directory discovery;
- apply exclusions consistently to background indexing and relevant file-watcher/on-demand indexing paths;
- keep explicit CLI file arguments checkable even if they match an exclusion;
- keep this separate from lint overrides and `.lintr` `exclusions`;
- document path matching rules, including directory globs and negation if supported.

## Performance notes

Sight's implementation has a useful performance shape that Raven should copy:

- compile glob patterns once into a matcher, not per path;
- provide an empty-config no-op fast path so normal scans pay almost nothing;
- match files and directories against paths relative to the containing workspace root;
- prune whole directories during the serial directory collection phase, before building the file list for parallel parsing;
- only prune directories for patterns that prove every descendant is excluded, e.g. globstar directory patterns;
- disable or conservatively limit directory pruning when negated re-include patterns are present;
- apply the same matcher to incremental/watch/on-demand indexing so newly excluded files drop stale entries.

Raven already depends on `globset`, so this likely does not need a new dependency. The important implementation detail is to integrate the matcher into `collect_files_matching` rather than filtering after all candidate paths have already been collected.

Relevant Sight implementation references:

- `src/utils/exclude-matcher.ts`
- `src/indexer/index.ts` directory pruning and incremental `index_file` exclusion
- `src/cli/source-files.ts` default walk exclusions while preserving explicit file targets

## Test coverage

Add tests for:

- excluded directories are skipped during workspace scan;
- excluded files are skipped by default `raven check` discovery;
- explicit file targets still run;
- config layering/recompute behavior updates exclusions correctly;
- directory pruning does not drop re-included files when negation is supported;
- exclusion filtering removes stale index entries on incremental updates.

jbearak:
Adds a project-level exclusion setting so large generated, vendored, output, or archived trees can be kept out of workspace indexing, dependency discovery, diagnostics, and default `raven check` discovery — separate from lint overrides and `.lintr` `exclusions`.

## What's new

```toml
[workspace]
exclude = ["generated/**", "archive/**", "!generated/keep.R"]
```

`raven.toml`-only (no VS Code/LSP client setting in this PR; that could be a follow-up if wanted). Default `[]`.

## Matching rules

- Ordered, project-root-relative globs matched against paths relative to the containing workspace root; **last matching pattern wins**.
- Leading `!` re-includes matching paths.
- Globs compile once (globset, already a dependency) with `literal_separator(true)`: `*`/`?` never cross `/`; use `**` for recursion (`generated/**`, `**/cache/**`).
- Empty config is a no-op fast path (`Cow::Borrowed` / `None` matcher everywhere).
- Directory pruning happens during the serial walk in `collect_files_matching`, before the parallel parse phase, and only for positive patterns that prove every descendant is excluded (`dir/**`-style). Any negated pattern disables pruning conservatively; the full ordered matcher still filters files.
- Explicit CLI **file** arguments bypass exclusions (`raven check generated/one.R` still reports it); **directory** arguments are discovery walks, so exclusions apply inside them.

## Where exclusions apply

Workspace scans (startup + `raven check` indexing), default `raven check` discovery, watched-file/incremental updates (newly excluded files drop stale index entries), on-demand indexing and disk resync, package-mode seeding (`R/`, `data/`, `data-raw/`, `.Rprofile` prelude incl. transitive `source()` helpers, watcher rescans), cross-file dependency edges (an excluded file lends no symbols even while open — identical to its closed behavior), and diagnostics publishing (excluded open docs publish an explicit empty list). Live `raven.toml` reload converges both directions: newly excluded entries are removed and force-republished; newly re-included files are rescanned (or, with `crossFile.indexWorkspace = false`, open buffers are replayed) — open-buffer topology always wins over disk truth. The compiled matcher is produced only in `recompute_parsed_configs`, per the config-layering invariant.

## Tests

All six required coverage bullets have tests, plus the review-round regressions (~20 new tests total), including:

- `workspace_scan_skips_project_excluded_directory`, `collect_targets_empty_skips_project_excluded_files`
- `explicit_file_target_bypasses_project_exclusion` (end-to-end through raven.toml)
- `recompute_updates_workspace_exclusions_from_project_layer`
- `collect_files_matching_negated_exclusion_does_not_prune_reincluded_file`, `last_match_wins_with_negation`, `single_star_does_not_cross_directory_separator`
- `watched_file_change_removes_stale_entry_when_now_project_excluded`
- reload convergence: `workspace_exclusion_reload_reindexes_newly_included_files`, `..._preserves_dirty_open_buffer_source_edge`, `..._reseeds_package_inputs_when_index_workspace_disabled`, `..._reincluded_open_helper_without_workspace_index`
- package seeding: `close_of_project_excluded_package_source_removes_package_input`, data/data-raw/.Rprofile exclusion tests
- open-file symbol lending: `open_project_excluded_sourced_helper_does_not_resolve_in_parent`, open/close no-flap, re-include restores resolution

## Review chain

codex implement → 6 codex adversarial review rounds (9 real findings found and fixed across 5 fix passes: did_close package re-seed leak; re-include on reload; `*` crossing `/`; open-buffer topology clobber on reload merge; indexWorkspace=false package reconciliation; data/data-raw/.Rprofile seeding; excluded-open-file symbol lending; indexWorkspace=false open-buffer replay) → codex clean → Sonnet second opinion clean (zero findings at confidence bar).

## Gates (all green on HEAD)

- `cargo fmt --all --check`
- `RUSTDOCFLAGS="-D warnings" cargo doc -p raven --no-deps --document-private-items` (default and `--features test-support`)
- `cargo test -p raven --features test-support` — 5675 passed; the 17 `cli::packages` failures in-sandbox are the known `TcpListener::bind` sandbox artifact and pass when run normally (verified: 51/51).

Deliberate scope-outs: `raven packages` referenced-package scanning ignores exclusions (over-inclusion is documented harmless there); excluded open buffers stay open and publish empty diagnostics for themselves.

Fixes #559

https://claude.ai/code/session_01VjnMh1eFxmXGR7dSc3o73M

## Summary by CodeRabbit

* **New Features**
  * Added `[workspace].exclude` in `raven.toml` to control which workspace paths are discovered and indexed, including `!` re-inclusion and directory pruning behavior.
  * `raven check` now applies these exclusions to what gets analyzed and indexed, while still diagnosing explicitly provided `PATHS`.
* **Bug Fixes**
  * Excluded files no longer seed related indexing, dependency graph metadata, diagnostics, or data/sysdata symbol generation.
  * Live changes to exclusions are picked up consistently.
* **Documentation**
  * Updated CLI and configuration docs to explain discovery/indexing vs diagnostic targeting and the glob semantics.


jbearak:
Thanks for the review. Addressed in the latest commit (62beaa1f):

**Finding 1 — `exclusions.rs` `relative_path` (raw `strip_prefix` vs symlinked/differently-spelled roots): fixed, but deliberately NOT via blanket `fs::canonicalize`.** We investigated where the paths fed to `is_excluded_path`/`can_prune_directory` originate:
- The serial workspace walk (`state.rs` `collect_files_matching_inner`), the on-disk package input collection (`backend.rs`), and the config-reload/reseed scans all build candidate paths by joining entries onto the *stored* workspace root, so root and path already share spelling — raw `strip_prefix` is authoritative there and needs no normalization.
- The one reachable gap is the watched/incremental path: `event.rs` `translate_watched` passes the raw `uri.to_file_path()` result (from the LSP client/watcher) into `is_excluded_path`, and that URI can arrive under a symlinked or differently-spelled root.

Rather than canonicalize unconditionally (this repo made a deliberate decision in #562/#565 to keep index/graph/cache keys uncanonicalized), `relative_path` now tries raw `strip_prefix` **first** (preserving the fast, zero-syscall serial-walk behavior) and only falls back to a canonicalized root-vs-input comparison on a raw miss. Canonicalization is confined to the matcher's input normalization — the `Cow<Path>` result is consumed solely for glob matching, and no key semantics change anywhere. Added a `#[cfg(unix)]` regression test (`symlink_spelled_watched_path_uses_canonical_fallback`) covering both existing and not-yet-existing leaf paths.

**Finding 2 — `docs/cli.md` "whole workspace is always indexed" contradiction: fixed.** Reworded the "Workspace and paths" overview to say the workspace is indexed except for `[workspace].exclude` paths, and clarified that explicit file arguments are reported as named while directory walks apply exclusions (consistent with the bypass note later in the file).

**Finding 3 — `event.rs` deleted-on-exclude comment: added.** Documented that a newly-excluded path is deliberately folded into `deleted` so every downstream cleanup branch reuses the same removal logic.

Verified locally: `cargo fmt --all --check`, `cargo clippy --workspace --all-targets --features test-support -- -D warnings`, `RUSTDOCFLAGS="-D warnings" cargo doc -p raven --no-deps --document-private-items` (default + `--features test-support`), and `cargo test -p raven --features test-support` (5693 passed, 0 failed) all green.

PR #597, opened 2026-07-09
object_name lint: support multiple styles (OR) and custom regexes

### LINKED ISSUE #512: object_name lint: support multiple styles (OR) and custom regexes
## Background

Raven's `object_name` lint ([`crates/raven/src/linting/rules/object_name.rs`](crates/raven/src/linting/rules/object_name.rs)) currently accepts exactly **one named style per symbol kind** (function / variable / argument), drawn from a fixed set: `snake_case`, `camelCase`, `dotted.case`, `UPPER_CASE`, `lowercase`, `any`.

As of #511, the `.lintr` reader handles `object_name_linter(...)` by mapping a single recognized style and **warning + ignoring** anything it can't represent — specifically:

- a **multi-style vector** like `object_name_linter(styles = c("snake_case", "camelCase"))` (lintr accepts a name matching *any* listed style — OR-semantics), and
- a **custom regex** like `object_name_linter(regexes = "^[a-z][a-z0-9_]*(\\.([a-z][a-z0-9_]*))*$")`.

Both are first-class in `lintr`. This issue tracks adding real support so these configs work instead of being dropped.

`lintr`'s actual signature is `object_name_linter(styles = "snake_case", regexes = character(0L))`: `styles` is the validated named-style set (OR), `regexes` is custom regexes (OR), and a name passes if it matches **any** accepted style **or** **any** regex. (Note: a raw regex passed positionally to `styles`, as in the literal example above, is actually invalid in real lintr — regexes belong in `regexes =`. Raven warning on that today is arguably *more* correct, but see the "lenient extension" option below.)

## Current architecture (for scoping)

- Config: three scalar fields on `LintConfig` — `object_name_style_{function,variable,argument}: ObjectNameStyle` ([`config.rs:80-88`](crates/raven/src/linting/config.rs), defaults `SnakeCase`). `ObjectNameStyle` is `#[derive(..., Copy, ..., Eq)]`; `LintConfig` is `#[derive(Debug, Clone, PartialEq)]`.
- Single source of truth: [`ObjectNameStyle::from_config_name`](crates/raven/src/linting/config.rs) (known name → variant, else `None`), consumed by the JSON path (`backend::parse_object_name_style`) and the `.lintr` loader (`parse_object_name_styles`).
- Carrier built **per lint pass** from the config at [`mod.rs:169`](crates/raven/src/linting/mod.rs) → `ObjectNameStyles { function, variable, argument }`.
- Matching: [`style_for(kind)`](crates/raven/src/linting/rules/object_name.rs) → `matches_scheme(name, style)` → per-style **byte-scan** predicates (`is_snake_case`, etc.). No regex today. A leading `.` is stripped before named matching; S3 generics are exempt.
- `regex = "1.10.0"` is **already a dependency** (used in `chunks.rs`, `handlers.rs`, `file_path_intellisense.rs`).

## Proposal

Model each kind as an **ordered list of acceptable patterns**, each either a named style or a regex; a name conforms iff it matches **any** of them. This unifies both features and mirrors lintr's `styles` ∪ `regexes`. The `.lintr` loader already parses `styles` into a `Vec<String>` and just needs to classify each element rather than gate on "single named only".

### A. Multiple named styles (OR-semantics)

- **Config:** widen `object_name_style_*: ObjectNameStyle` → `Vec<ObjectNameStyle>` (default `[SnakeCase]`). `Vec<ObjectNameStyle>` keeps `LintConfig`'s `Debug/Clone/PartialEq` derives intact (no friction). `Any` anywhere in the list short-circuits to "no check for this kind".
- **Carrier/checker:** `ObjectNameStyles` fields become `Vec<ObjectNameStyle>` (or `&[…]`); `matches_scheme` → `styles.iter().any(...)`.
- **Message:** when nothing matches, list the accepted styles ("…does not match any of: snake_case, camelCase").
- **Config plumbing:**
  - LSP/`raven.toml` JSON: accept a string **or** an array of strings for `objectNameStyle*` (backward compatible — string still works).
  - `.lintr`: emit `objectNameStyleFunction` as an array; drop the multi-style "unrecognized construct" warning.
  - VS Code `package.json`: `objectNameStyle*` schema becomes `oneOf: [<enum string>, array of <enum string>]` (still validated).

**Effort: Medium (~1 day).** Type widening + message + string|array parsing + docs + tests at all layers + the settings-sync invariant (`package.json`, `initializationOptions.ts`, `SETTINGS_MAPPING`, `settings.test.ts`, regen `settings-reference`). No new concepts, no derive friction.

### B. Custom regexes

- **Config:** add `object_name_regexes_{function,variable,argument}: Vec<String>` (store the **source strings**, not `regex::Regex` — `Regex` is neither `Eq` nor `PartialEq`, which would break `LintConfig`'s derives). Compile to `Regex` when building the carrier at [`mod.rs:169`](crates/raven/src/linting/mod.rs) (the carrier is short-lived and need not be `PartialEq`).
- **Validation:** attempt compilation at config-parse time and `log::warn!` + skip on invalid patterns, so bad regexes surface once and never panic the linter.
- **Matching semantics (decisions to lock in):**
  1. **Partial vs anchored:** mirror lintr's `grepl` → `Regex::is_match` (partial); anchoring is the user's responsibility. Document this.
  2. **Leading-dot / S3 handling:** apply regexes to the **full identifier** (not the dot-stripped body the named matchers use), so patterns like the example that explicitly encode dotted segments work. Keep the S3-generic exemption.
  3. A name passes if it matches any named style **or** any regex.
- **Config plumbing:**
  - Prefer **separate settings** `objectNameRegexes{Function,Variable,Argument}: string[]` (mirrors lintr's `styles`/`regexes` split) — this keeps the `objectNameStyle*` enum **dropdown** in the VS Code UI instead of degrading it to free-text. Each new setting needs the full settings-sync wiring.
  - `.lintr`: parse `object_name_linter(regexes = c(...))` into `objectNameRegexes*`. **Lenient extension (recommended):** also classify any element of `styles` that isn't a known name as a regex (handles the user's literal positional-regex example, a superset of lintr) — call this out in `docs/linting.md`.
  - **Perf:** compiling a handful of small regexes per lint pass is likely negligible, but the carrier is rebuilt on every edit; if benchmarks show cost, cache compiled patterns alongside the resolved config.

**Effort: Medium-Large (~2-3 days).** The LOC is modest; the cost is the decisions (partial-vs-anchored, dot handling, where to compile/cache), the invalid-regex warning path, three new settings × the settings-sync invariant, the VS Code UX trade-off, docs, and tests.

### Recommendation

Do **A and B together** as one "acceptable patterns" model. The `.lintr` loader already yields a `Vec<String>`; doing both at once means classifying each element once (known name → style; else → regex) instead of migrating the same three fields twice.

## Test plan

- Config layer: array parsing (string-or-array), regex compilation + invalid-regex warning, defaults.
- `.lintr` loader: `styles = c("a","b")` → array (no warning); `regexes = c(...)`; mixed `styles` with a regex element; the user's positional-regex example now maps instead of warning.
- Checker: OR across named styles; OR across regexes; named ∪ regex; anchored vs unanchored behavior; leading-dot and S3 cases; multi-style "does not match any of" message.
- End-to-end `.lintr` → `LintConfig` → diagnostics (extend the patterns added in #511).
- VS Code settings round-trip + `settings-reference` drift test.

## Open questions

1. Faithful lintr (`regexes` only) vs the lenient extension (regex allowed in `styles` too)? Recommended: support both, document the extension.
2. Regex match: partial (`grepl`-like, recommended) or force full-match anchoring?
3. Apply regex to the full name or the dot-stripped body? Recommended: full name.
4. Keep the `objectNameStyle*` enum dropdown by adding separate `objectNameRegexes*` settings (recommended), or collapse to free-text string|array?

Refs: #511 (added the warn-and-ignore behavior this would supersede), #220 (original `object_name_linter`).

jbearak:
Implements #512: the `object_name` lint now models each symbol kind (function / variable / argument) as an OR-list of accepted patterns — named styles plus custom regexes — mirroring lintr's `object_name_linter(styles = ..., regexes = ...)`. A name conforms iff it matches **any** accepted named style **or** **any** regex.

## Server

- `LintConfig`: `object_name_style_*` widen to `Vec<ObjectNameStyle>` (default `[snake_case]`); new `object_name_regexes_*: Vec<CompiledRegex>` compiled once at config-parse time (never per lint pass), with `Debug`/`PartialEq` by source string so the struct derives survive.
- Matching: regexes are **partial** matches (grepl-like) against the **full identifier** (leading dot intact); named styles keep the dot-stripped-body semantics. `any` anywhere in a styles list disables that kind; explicit empty styles + regexes = regex-only; both empty = disabled. S3-generic / backtick / non-ASCII carve-outs unchanged.
- JSON / `raven.toml`: `objectNameStyle*` accept a string **or** array (backward compatible — absent keeps the default, explicit `[]` is regex-only, non-empty-but-all-invalid fails safe to `[any]` with a warning). New `objectNameRegexes*` string-array keys with compile-time validation; empty patterns are rejected (an empty regex matches every name and would silently disable the lint).
- `.lintr`: multi-style vectors and `regexes =` now map instead of warning (supersedes the #511 warn-and-ignore behavior). Unknown non-empty `styles` elements are leniently treated as regexes — a documented Raven extension — with a did-you-mean warning for case/punctuation typos of known style names. R string escapes are decoded (`"^\\.on[A-Z]"` → `^\.on[A-Z]`) and R raw strings (`r"(...)"`, dash-extended and bracket/brace forms) are supported. Malformed tokens (bare identifiers, adjacent literals, quoted-empty elements, all-empty vectors like `c(,)`) count toward the batch unrecognized-construct warning. Explicit `c()` for `styles` or `regexes` emits `[]` so a project `.lintr` overrides client-layer values through the config merge — the same meaning the equivalent `raven.toml` has.

## VS Code

- `objectNameStyle*` schema becomes `oneOf` string-or-array (enum dropdown preserved); three new `objectNameRegexes*` array settings wired end-to-end: `initializationOptions.ts`, `SETTINGS_MAPPING` (new enum-or-array harness type with scalar/array/empty-array coverage), `renderRavenToml` + `.lintr` scaffold with key-aware array validation, `KNOWN_LINTING_KEYS`, and the regenerated settings reference.

## Docs

`docs/linting.md`, `docs/configuration.md`, `docs/diagnostics.md`, `docs/settings-reference.md` (regenerated).

## Review process

Plan adversarially reviewed by two independent reviewers before implementation; the diff then went through eight review rounds (paired independent finders + adversarial verification per candidate), fixing among others: R-string escape handling, an empty-regex silently-disables-lint trap, octal-escape corruption, raw-string support, a cross-layer `regexes = c()` override gap, and a `check_parameters` short-circuit that would have broken regex-only argument checks. The final two rounds came back clean.

## Test plan

- `cargo fmt` / `clippy -D warnings` / `cargo test -p raven` (5821 pass) / rustdoc `-D warnings` in both feature configs — all green.
- `bun test` (1493 pass, includes the VS Code integration suite and both settings-reference drift gates).
- New coverage at every layer: style-list OR semantics, regex-only kinds, disabled-predicate edges, string-vs-array parsing, invalid/empty regex rejection, `.lintr` forms (multi-style, positional regex, raw strings, escapes, explicit-empty vs malformed-empty vectors), settings round-trips, scaffold key-aware arrays.

Closes #512

https://claude.ai/code/session_01U19cMwSwr8Tiw9ghpATyhG

## Summary by CodeRabbit

* **New Features**
  * Added per-kind object-name matching that combines configured naming styles with additional Rust-regexes for functions, variables, and arguments.
  * Updated editor/config settings so `objectNameStyle*` accepts either a single style or an array; added `objectNameRegexes*` array settings.
* **Bug Fixes**
  * Corrected “disabled” vs “regex-only” behavior, including explicit empty style/regex arrays and safer fallbacks when regexes are empty/invalid.
  * Empty/invalid regex entries are ignored; unknown styles warn and are skipped.
* **Documentation**
  * Refreshed object-name linting docs, examples, and settings reference for the style+regex model.
* **Tests**
  * Expanded parsing and integration coverage for scalar/array and regex-only/disabled semantics.


jbearak:
@coderabbit review

Issue #610, opened 2026-07-11
Per-axis indentation settings (aligned/indented/off), lint default to 'either', and a docs coherence pass


jbearak:
## Motivation

#606 gave the **linter** a configurable infix-continuation style (`raven.linting.infixContinuationStyle`: `indented`/`aligned`/`either`), but auto-indent has no counterpart: the smart indenter's chain-start alignment is unconditional — `calculator.rs`'s `AfterContinuationOperator` branch computes `max(chain_start_col, line_indent + tab_size)` and never consults `config.style`. Users who prefer the RStudio/lintr one-level continuation style cannot get the indenter to produce it without turning Tier 2 off entirely.

Separately, `raven.indentation.style`'s ESS-derived value names (`rstudio`/`rstudio-minus`) read as whole-editor behavior bundles — which is what a *style* means in ESS — but the setting only governs paren-argument alignment. In ESS, the `RStudio` style is genuinely RStudio-faithful across all axes (it even sets `ess-align-continuations-in-calls` to nil); Raven's `rstudio` value aligns operator chains, which the real RStudio IDE never does (it indents statement-level continuations one level — verified against `getNextLineIndent` in rstudio's `r_code_model.js`). This mismatch caused real confusion and stale docs (fixed in #609).

## Proposal

Two per-axis settings, shared vocabulary, each with `aligned` / `indented` / `off`:

- **`raven.indentation.argumentStyle`** — paren-argument treatment.
  - `aligned` (default) — same-line args align to the column after the opener (= today's `rstudio`)
  - `indented` — one level from the opener's line (= today's `rstudio-minus`)
  - `off` — Tier 2 stands down for this construct; Tier 1 stands
- **`raven.indentation.infixContinuationStyle`** — operator-chain continuations.
  - `aligned` (default) — chain-start alignment with the one-level floor (today's behavior)
  - `indented` — uniform one level anchored at the chain-start line, matching the RStudio IDE and lintr's default (NOT Tier 1's per-line cascade)
  - `off` — Tier 2 stands down for this construct; Tier 1 stands

No `either` on the indent side — a producer must pick one form; `either` only makes sense for the lint judge. The leaf name deliberately mirrors `raven.linting.infixContinuationStyle` so the producer/judge symmetry is discoverable, and `indented`+`indented` gives RStudio-style users symmetric behavior on both sides.

[...]

PR #612, opened 2026-07-12
Judge-backed on-type indentation: assignment continuations indent one level (#611)

### LINKED ISSUE #611: Auto-indent puts the continuation of a line ending in an assignment operator at column 0
## Bug

Pressing Enter after a line ending in an assignment operator leaves the next line at column 0:

```r
result <-
f(x)        # cursor lands at column 0; should be one indent level
```

Every reference point says one indent level, and Raven's own linter agrees:

- **Raven's indentation lint flags Raven's own auto-indent output** — `result <-` ⏎ `f(x)` at column 0 produces `Indentation should be 2 spaces, not 0.`, while the 2-space form is clean (verified through `run_lints` with an indentation-only config). This violates the documented invariant that the linter never flags indentation its own auto-indent just produced.
- **Real lintr** (3.3.0.1, `indentation_linter()` defaults) flags the same shape identically: `Indentation should be 2 spaces but is 0 spaces.`
- **RStudio and ESS** both indent one level after an assignment-operator-ended line (ESS documents that definition operators "still trigger an indentation in all cases").

## Cause

Two independent gaps:

1. **Tier 2**: `detect_context` classifies the incomplete `result <-` as `AfterCompleteExpression` → column 0. Assignment operators are not in the continuation-operator list (`|>`, `%>%`, `+`, `~`, `%…%`), so no continuation context fires.
2. **Tier 1**: the `onEnterRules` in `editors/vscode/language-configuration.json` cover the same operator list and also skip assignment operators, so nothing catches it even with Tier 2 off.

## Design decision (maintainer)

A line ending in an assignment operator indents **one level, in every style** — including the future indent-side `aligned` mode from #610 — because there is nothing to align against: the RHS of an assignment is not a peer operand. This mirrors the lint side, where #606 deliberately exempts assignment operators (`<-`, `<<-`, `=`, `:=`, `->`, `->>`) from `infixContinuationStyle` in every mode. Producer and judge stay symmetric.

## Fix shape

- Tier 2: detect assignment-operator-ended lines as a continuation context that indents one level from the assignment line's indent (`line_indent + tab_size`), unconditionally across styles.
- Consider which operators participate: `<-` / `<<-` are clear; `=` needs the AST to distinguish top-level assignment from a named argument ending a line inside a call (where `InsideParens` handling should win); `->` / `->>` -ended lines are rare but valid and can use the same rule.
- Tier 1 (optional belt-and-suspenders): an `onEnterRule` for `<-` / `<<-`-terminated lines. `=` should stay Tier 2-only since regex cannot disambiguate it.

## Verification requirements (per #610's discipline)

- Pin the fixed behavior with tests through the real `detect_context` → `calculate_indentation` pipeline, including the follow-on case: after the fix, a chain starting on the indented line (`result <-` ⏎ `data %>%` at one level) gets continuations at `max(chain_start_col, chain_line_indent + tab)` — verify the default lint accepts the resulting shape before documenting it as the long-LHS escape hatch discussed in #610.
- Update `docs/indentation.md` (this shape is currently undocumented) and re-verify every documented column against the engine.

Found while researching #610: https://claude.ai/code/session_0153TLP6mjTRpd5JJJZkovkV

jbearak:
Closes #611.

Pressing Enter after a line ending in an assignment operator (`<-`, `<<-`, `=`, `:=`, `->`, `->>`) now indents one level instead of landing at column 0, in every style, matching RStudio/ESS/lintr and Raven's own indentation lint.

## Architecture

Tier 2 on-type indentation is **judge-or-nothing**: the handler repairs the just-edited buffer (sentinel + synthesized closers), incrementally reparses, and asks the indentation lint's own expectation engine (`accepted_indents_for_lines`) for the accepted column set, selecting by style preference. Every emitted column is lint-accepted *by construction*. When the judge cannot answer (multiline-string interiors, tabs-mode, unrepairable error windows, non-conforming reference lines), the server emits **no edit** and the editor's Tier 1/native indentation stands.

The legacy `detect_context`/`calculate_indentation` fallback engine — a complete parallel indenter with its own string/raw-string/comment scanners — is deleted (~8,700 lines net). Its bail-class behavior deltas were enumerated by differential replay and are pinned in `docs/indentation.md`.

- Tier 1 gains a declarative `onEnterRule` for `<-`/`<<-` (R config + generated Rmd/Quarto variant so the rule cannot fire in prose); `=` is Tier 2-only since a regex cannot distinguish named arguments.
- `isIndentUnitDocument` honors `languageId` for file-backed documents; indent-unit document sets are derived from the R-document sets by construction.
- New criterion bench `on_type_indentation/enter_bottom_10000_lines` (~4.2 ms measured; budget p95 ≤ 16 ms at 10k lines) and a declined-findings registry in `docs/development.md`.

## Review history

Six adversarial review rounds (finder: Codex sol xhigh; verify: Opus) plus two independent architecture assessments converged on this design; a final review round on the deletion commit surfaced one PLAUSIBLE finding (tabs-mode Tier 2 stand-down), which is the deliberate, documented delta class and is now pinned in the registry.

Follow-ups (approved plan): PR 2 ships the combined #610 settings model (per-axis `argumentStyle`/`infixContinuationStyle`, lint default `either`); PR 3 adds the settings-mismatch advice diagnostic.

https://claude.ai/code/session_013A7tqY8fJs8WDn9Ti45g2M

## Summary by CodeRabbit

* **New Features**
  * Added AST-aware smart indentation for Enter, with support for configured indentation styles and continuation layouts.
  * Added improved indentation handling for R Markdown and Quarto documents in VS Code.
  * Added assignment-related Enter indentation rules.

* **Bug Fixes**
  * Improved per-document indentation settings and configuration updates.
  * Safely preserves existing editor indentation when Raven cannot determine a reliable result.

* **Documentation**
  * Updated smart indentation and continuation-style guidance.

* **Tests**
  * Added broad coverage for indentation behavior, editor configuration, and performance.


docs/linting.md
```diff
@@ -270,7 +270,7 @@ Each rule lists the Raven settings that control it and the `lintr` linter it mir
     )
     ```
 
-    This is a strict requirement, not an auto-indent-parity mode: [smart indentation](indentation.md) indents a continuation to at least one unit, so for a chain starting at a line's first column (e.g. a top-level `data |>` pipeline) auto-indent suggests one unit while `"aligned"` demands column 0. It also disables the aligned-argument/block-form tolerances on the specific lines an operator continuation covers.
+    This is a strict requirement. [Smart indentation](indentation.md) queries the configured lint style before choosing a column, so with `"aligned"` a top-level `data |>` pipeline continues at column 0 as required; auto-indent does not emit the otherwise-preferred one-unit form and then get flagged. This mode also disables the aligned-argument/block-form tolerances on the specific lines an operator continuation covers.
```
jbearak, line 273: This isn't what "aligned" was supposed to mean. You'd never want:

```
y <-
x
```

or

```
x |> 
y
```

What "aligned" gets you is:

```
x <- y |>
        z
```

whereas "indented" gets you:

```
x <- y |>
    z
```

Supposing the user chooses the "aligned" style and types:

```
x <- y +
     z + 
     w
```

If however they type:
```
x <-
    y +
    z
    w
```

"y" and "z" and "w" are all aligned. 

It's ifne in this case that the "x" isn't aligned with "y" and "z" and "w". That's not what aligned is supposed to be about.

jbearak: Confirmed and resolved in the follow-up: `aligned` means an RHS begun on the next line gets one indent level (the assignment rule, in every mode), and subsequent chain operands then align with that first operand's actual column — never with the LHS, and never column 0. The sentence this comment anchors to documents the pre-existing #606 strict behavior; the next PR (combined #610 settings model) replaces that behavior with the floored definition and rewrites this section, with your four examples pinned as tests.

PR #649, opened 2026-07-16
feat(cross-file): computed source() paths + testthat working-directory convention (#638)

### LINKED ISSUE #638: Support testthat working-directory convention + computed source() paths for non-package suites
## Summary

Raven flags top-level definitions as undefined-variable in a `testthat` suite when those definitions come from project code that a `helper-*.R` file pulls in via a **computed `source()` path** anchored at the testthat working directory. This is common in **non-package** projects (code under `scripts/`, `src/`, etc. rather than `R/`), where the standard idiom is to compute the repo root from testthat's working directory and source project files from there.

Raven already supports the core testthat helper convention (helper-*.R defs visible to peer test-*.R files, #275). The gap is that it cannot follow the `source()` calls the helper itself makes, because (a) the path is a computed expression rather than a string literal, and (b) resolving it correctly depends on the testthat working-directory convention.

## Reproduction

Non-package project (has `DESCRIPTION` but no `R/`; functions live under `scripts/`):

```
DESCRIPTION
scripts/helpers.R                 # defines my_theme(), make_result()
tests/testthat.R
tests/testthat/helper-project.R
tests/testthat/test-contract.R
```

`tests/testthat/helper-project.R`:

```r
# testthat sources helper-*.R before test-*.R, and evaluates them with
# tests/testthat as the working directory.
repo_root <- normalizePath(file.path("..", ".."))
source(file.path(repo_root, "scripts/helpers.R"))
```

`tests/testthat/test-contract.R`:

```r
testthat::test_that("uses project helpers", {
    p <- my_theme()          # `my_theme` flagged: undefined-variable
    r <- make_result()       # `make_result` flagged: undefined-variable
})
```

Every symbol defined in `scripts/helpers.R` is flagged as undefined in the test file, even though the suite runs correctly. A symbol defined **directly** in `helper-project.R` is (correctly) *not* flagged — confirming the boundary is the un-followed `source()` edge, not the helper convention itself.

## Root cause

Two interacting limitations:

1. **`source()` path must be a string literal.** `find_file_argument` in `crates/raven/src/cross_file/source_detect.rs` extracts the path only via `extract_string_literal` (the sole non-literal form accepted is a `system.file(...)` call with literal parts). `source(file.path(repo_root, "scripts/helpers.R"))` is a `call` node with a variable base, so `try_parse_source_call` returns `None` and no forward-source edge is created.

2. **No testthat working-directory convention.** Raven resolves `source()` paths relative to the sourcing file's own directory (backend.rs: "child resolves its own source() paths relative to its own [dir]") and only overrides via the `# raven: cd` directive. testthat, however, evaluates helper/test files with `tests/testthat/` as the working directory (`test_dir()` / `test_local()`), which is exactly why the idiom computes `repo_root <- normalizePath(file.path("..", ".."))`. Even if the computed path were recognized, it must be anchored at the testthat CWD, not the file's directory.

## Proposed support

For files under `tests/testthat/`:

- **Apply testthat's working-directory convention:** treat the working directory as `tests/testthat/` when resolving relative paths and `source()` targets (matching `test_dir()`/`test_local()` runtime behavior). This should be automatic for the standard layout, not require a `# raven: cd` directive.
- **Statically fold the common path expressions** so the computed `source()` target resolves: `file.path("a", "b", ...)` with literal parts, `normalizePath(...)` wrapping such an expression, and a local variable bound to one (`repo_root <- normalizePath(file.path("..", ".."))` → then `file.path(repo_root, "scripts/helpers.R")`). Non-static components can still bail out.

Together these let Raven follow the helper's `source()` edge and pick up the project definitions, eliminating the false undefined-variable diagnostics in non-package testthat suites.

## Workarounds (for reference)

- Declare the edge explicitly: `# raven: source ../../scripts/helpers.R` in the helper.
- Replace the computed path with a literal relative path: `source("../../scripts/helpers.R")` (runtime-equivalent given testthat's CWD).

Both are repo-side; native support is preferable because the computed-path idiom is idiomatic for non-package testthat projects.

## Related

- #275 (helper-*.R defs visible to peer test files) — the convention this builds on.

jbearak:
Support the idiomatic non-package testthat pattern where a helper computes
the repo root and sources project code through it:

  repo_root <- normalizePath(file.path("..", ".."))
  source(file.path(repo_root, "scripts/helpers.R"))

Three pieces:

1. Static path folding (cross_file/static_path.rs): fold file.path() of
   statically-known parts, normalizePath() wrappers, and single-assignment
   top-level variables into plain paths at detection time. Strict
   all-or-nothing (non-fsep named args, escapes, reassigned/conditional/
   scoped bindings all bail). Shared by source detection and go-to-def so
   the surfaces cannot drift; .Rprofile prelude scanning follows folded
   paths too.

2. Implicit testthat/testit working directory (path_resolve.rs): files
   under tests/testthat|testit/ anchor forward-relative paths at that
   directory, mirroring test_dir()/test_local(). Layout-only gating, soft
   default: excluded from cd_in_effect() (workspace fallback stays active,
   nothing propagates to sourced children), overridden by # raven: cd,
   with a file-directory compatibility fallback for nested files. Applied
   uniformly across edges, scope, diagnostics, go-to-def, and completion
   (metadata-less forward fallbacks now use forward_without_metadata).

3. Preamble sourced-closure scan (package_state/preamble.rs): harvest
   top-level defs and library() attaches from files a helper*/setup*.R
   preamble transitively sources, merged into test_helper_symbols at
   derive time so sibling test-*.R files see them. Mirrors the .Rprofile
   prelude subsystem: off-lock disk scan, watched-file freshness rescans,
   bounded worklist, suppressive-only.

Go-to-definition now resolves the full joined path from any segment of a
computed source() expression (previously it navigated to the single
literal under the cursor), and path completion in testthat files offers
entries from the anchor directory.

Closes #638

Claude-Session: https://claude.ai/code/session_01HwYMbxB7bHuPeMVNLX8tt8

## Summary by CodeRabbit

* **New Features**
  * Added strict static folding for computed `source()`/`sys.source()` paths (e.g., `file.path()` and `normalizePath()`), improving symbol resolution and diagnostics, including go-to-definition and path completion.
  * Introduced an implicit working-directory anchor for files under `tests/testthat/` and `tests/testit/`.
  * In package mode, test preambles now track and refresh sourced-closure dependencies, including live edits and restoration after closing.
* **Bug Fixes**
  * Ensured the implicit anchor is not inherited through sourced child files.
* **Documentation**
  * Updated cross-file `source()` rules and test and package-dev behavior.
* **Tests**
  * Added computed-path and live preamble refresh coverage, plus updated related fixtures.


jbearak:
@coderabbitai review

Issue #656, opened 2026-07-17
Follow-up simplifications discovered during PR #655 review


jbearak:
## Summary

PR #655's review identified several optional, behavior-preserving simplifications. They are deferred to keep that already-large patch focused on correctness and to avoid restarting its review surface.

These are **cleanup candidates, not known bugs**, and do **not gate PR #655**. Each item must be revalidated against the then-current code before implementation. Prefer small subsystem-specific PRs rather than landing this checklist as one patch.

## Reassess after #650

These overlap with or may be reshaped by the analysis-store consolidation in #650:

- [x] Simplify `ContentProvider::exists_cached()` by reusing the authoritative open-document predicate. #650 intends to reduce `ContentProvider` to a thin view over one store, so the current lookup structure may disappear.
- [x] Reassess delegating metadata-free artifact construction to the metadata-aware builder against #650's single derived-data chokepoint. The duplication is real, but #650 may determine the correct abstraction boundary.

## Independent cleanup candidates

These are not materially dependent on #650 and may be handled before or after it in focused PRs.

### Package state

- [x] Consolidate the duplicated recursive package R-file scanners in `package_state/event.rs`; empty compiled exclusions already provide no-op filtering.
- [x] Avoid constructing the incremental preamble `sourced_files` union that `apply_rescanned_preambles` immediately discards and recomputes from live keyed state.
- [x] Share cached-set merging in `package_state/derive.rs` and preserve the existing `Arc` whenever the added set is already a subset.
- [x] Skip routing canonicalization in preamble source reads when the override map is empty.

### Dependency graph

- [x] Remove the redundant `edges_changed` side of the cache-revision condition after revalidating that `EdgeRevisionIdentity` strictly contains the complete dependency-interface identity plus lending policy.

### Scope analysis

- [x] Build the complete scope-event timeline before sorting, rebuilding the function-scope tree, and annotating function scopes, avoiding repeated intermediate passes.

### Binding analysis

- [x] Represent resolved `assign()` values as required rather than `Option`; argument resolution already rejects omitted and missing values.
- [x] Record binding kill events directly instead of appending and immediately popping a shadow event.
- [x] Remove the unused `persistent_uncertainty` write for the unknown-loaded-binding sentinel if its consumers still inspect only ordered shadow events.

### Tests

- [x] Remove the consecutive duplicate `SourceLocality::CurrentFrame` assignment in the revalidation test if it remains after #655 lands.

## Implementation guidance

[...]

PR #670, opened 2026-07-19
Simplify binding event recording

jbearak:
Refs #656; does not close it.

## Summary

This is the next bounded #656 subsystem slice. It keeps binding-analysis behavior unchanged while making three internal invariants explicit:

- make `ResolvedAssignArguments.value` a required syntax node because strict `assign()` matching already rejects omitted and explicitly missing `value` actuals;
- record shadow, definite-kill, and non-restoring-kill events directly through one shared accountant instead of temporarily appending and removing a shadow event;
- remove the unread `persistent_uncertainty` write from the unknown-loaded-name sentinel while preserving its scoped ordered shadow event and all helper/named uncertainty behavior.

The diff is intentionally limited to `crates/raven/src/cross_file/binding.rs`. No user-visible behavior or documentation changes are intended.

## Semantic parity

The shared event recorder preserves the prior generation materialization, saturating count increment, earliest `first_offset`, candidate and safe-eager fields, and scope-specific event ordering:

- `Shadow` appends one shadow event;
- `DefiniteKill` appends one kill event and can restore an alias after an earlier shadow;
- `NonRestoringKill` appends neither event, so an earlier possible shadow remains while unique-candidate accounting is still invalidated.

The only representation difference is that a kill-only scope has no shadow-map entry instead of an empty vector; every query treats those states identically.

The loaded-name sentinel continues to record count, generation, first offset, scope, and ordered shadow offset. Its removed persistent bit was never read by loaded-name queries; the corresponding helper and unknown-named persistent writes remain intact.

## Review

- The three candidates were independently traced against exact main before implementation.
- Fable reviewed the combined architecture read-only and recommended all three changes, the shared private event helper, and the semantic variant names used here.
- An independent post-edit regression/new-bug review found no blocker and rechecked every caller, accounting field, uncertainty query, and the corrected test fixtures.

## Validation

- focused binding module: 25 passed
- focused trusted-removal, uncertain-removal, loader-destination, and loader-side-effect parity tests: 4 passed
- deterministic event-state coverage uses a modeled multi-event sequence rather than repeated stress, per Fable's recommendation
- `cargo test -p raven --features test-support`: 6485 passed, 12 ignored; all integration suites and 53 doctests passed
- `cargo fmt --all --check`
- `RUSTDOCFLAGS="-D warnings" cargo doc -p raven --no-deps --document-private-items`
- `RUSTDOCFLAGS="-D warnings" cargo doc -p raven --no-deps --document-private-items --features test-support`
- `git diff --check`

## Base

Rebased with identical patch onto stabilized main commit `3de21dac8ca44d9672feb8a97da97a1b4a983ed2` after prerequisite PR #671.

### DIFF
--- a/crates/raven/src/cross_file/binding.rs
+++ b/crates/raven/src/cross_file/binding.rs
@@ -495,6 +495,29 @@ fn bump_at_in_scope<'m, T>(
     node: Node,
     scope: Option<LexicalScope>,
 ) -> &'m mut Binding<T> {
+    record_binding_event_at_in_scope(collection, name, node, scope, BindingEvent::Shadow)
+}
+
+#[derive(Clone, Copy)]
+enum BindingEvent {
+    Shadow,
+    DefiniteKill,
+    NonRestoringKill,
+}
+
+/// Record one binding event while applying the shared generation, count, and
+/// first-offset accounting exactly once.
+///
+/// A shadow participates in ordered alias invalidation, a definite kill can
+/// restore a base alias after an earlier shadow, and a non-restoring kill only
+/// invalidates unique-candidate accounting because its branch may not run.
+fn record_binding_event_at_in_scope<'m, T>(
+    collection: &'m mut BindingCollection<T>,
+    name: &str,
+    node: Node,
+    scope: Option<LexicalScope>,
+    event: BindingEvent,
+) -> &'m mut Binding<T> {
     let offset = node.start_byte();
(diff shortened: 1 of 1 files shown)


jbearak:
@coderabbitai review

jbearak:
@coderabbitai review

jbearak:
@coderabbitai full review

PR #683, opened 2026-07-20
Diagnose packages hidden by active renv

jbearak:
## Summary

- detect when an active renv project hides a package that is available in the user's other R libraries
- report `package-outside-active-library` with `renv::restore()` / project-install guidance
- preserve normal package routing: outside libraries contribute validated package names only, never exports, completions, resolution, watchers, or `system.file()` paths
- fail closed for uncertain activation and multi-root LSP sessions
- document CLI, diagnostic, suppression, and lifecycle behavior

## Root cause

Raven previously saw only the post-activation `.libPaths()` result. A checkout with `renv/activate.R` could therefore make a machine-installed package disappear without leaving enough information to distinguish it from a genuinely uninstalled package.

The existing bounded R discovery invocation now frames the before/after library transition and verifies the activated renv project. Raven reduces removed libraries immediately to validated package names and attaches that names-only overlay to the matching package-library lifecycle.

## User impact

`raven check` now explains that a referenced package exists outside the active renv project and suggests restoring or installing it into the project library. The CLI keeps the generic `package-not-installed` diagnostic suppressed by default while leaving this actionable child diagnostic enabled. Direct and umbrella suppression remain available.

## Validation

- synthetic, machine-independent probe/parser, package-overlay, diagnostic, suppression, CLI, lifecycle, and multi-root regressions
- `cargo test --workspace --all-targets --features test-support`
- `cargo fmt --all --check`
- rustdoc with warnings denied, default and `test-support`
- two independent agent reviews, both approved
- committed-patch audit found no machine paths or external-checkout material
- exact local release A/B: warm median full-check time was 1.18s on both current main and this commit

## Summary by CodeRabbit

* **New Features**
  * Added a `package-outside-active-library` missing-package diagnostic for packages present only outside the active `renv` library.
  * Provides actionable `renv::restore()` / install guidance and reports at most once per package per document.
  * Improves accuracy by avoiding related “missing member” noise when the package is only outside the active `renv` library.
* **Bug Fixes**
  * CI/default now suppresses generic missing-package diagnostics while keeping the new subtype enabled.
  * Multi-root and invalid `renv` evidence now fail closed (no subtype reporting).
* **Documentation**
  * Updated CLI and diagnostics docs to reflect the new subtype and `--report-uninstalled` behavior.
* **Tests**
  * Expanded coverage for suppression, deduping, and multi-root fail-closed behavior.

### DIFF
--- a/crates/raven/src/cli/check.rs
+++ b/crates/raven/src/cli/check.rs
@@ -476,12 +476,12 @@ async fn run_with_cwd(args: CheckArgs, cwd: &Path) -> i32 {
         Err(code) => return code,
     };
 
-    // CI default: suppress the missing-package ("not installed") diagnostic,
-    // because CI deliberately omits installation (spec §10.1). The CLI owns
-    // `state` exclusively, so a direct field set here is safe.
-    // `--report-uninstalled` opts back in.
+    // CI default: suppress generic missing-package diagnostics because CI
+    // deliberately omits installation (spec §10.1). Actionable subtypes such
+    // as "available outside this active renv project" remain enabled.
+    // `--report-uninstalled` opts the generic diagnostic back in.
     if !args.report_uninstalled {
-        state.cross_file_config.packages_missing_package_severity = None;
+        state.cross_file_config.packages_report_generic_uninstalled = false;
     }
 
     // Auto-detect R for installed-package / base-symbol awareness. Any failure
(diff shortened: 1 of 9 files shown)


jbearak:
Rebased onto merged stabilization main at 105eb5034589548c899229cfbb725ce8e2b476d4. The original implementation patch remains byte-equivalent under stable patch ID 3018683b355e9467b4bb192b4903b1561cf90aa5 (range-diff `=`). A separate follow-up commit converts the existing multi-root fail-closed invariant into the requested rustdoc. Fresh CI and review are required. @coderabbitai full review

PR #673, opened 2026-07-19
Stabilize close resync and package overlay tests

jbearak:
## Summary

- isolate `close_resync_overrides_stale_unified_index_entry` from detached startup scanning/routing and observe its exact disk-resync invocation through the existing typed final-handoff capture
- snapshot the fresh closed index and dependent diagnostics before releasing the capture, with a test-specific 30-second resync wait for process-global semaphore contention while preserving the existing 5-second default elsewhere
- disable unrelated startup package routing in the manually seeded sourced-helper OpenEdit fixture while preserving its package projection, routing-owner, reservation-cap, and symbol assertions

## CI evidence

This addresses both parallel-harness Cargo failures exposed by #672:

- attempt 1: `close_resync_overrides_stale_unified_index_entry` timed out while inferring an unjoined close resync from eventual shared state
- retry: `did_open_dirty_sourced_helper_overlays_rprofile_and_preamble_scans` observed routing owner `177 -> 178` from unrelated deferred package routing

The exact paths were revalidated as test fixture/observation races, not production behavior bugs. The change is test-only in `backend.rs`; no sleeps or production changes.

## Validation

- both focused tests: pass
- 100 repetitions per focused test: 200/200 pass
- `project_config_initialize_tests` with `--test-threads=32`: 328/328 pass
- two full `cargo test --workspace --all-targets --features test-support` runs: each 6,687 passed, 15 ignored, 0 failed
- `cargo fmt --all --check`
- `cargo clippy --workspace --all-targets --features test-support -- -D warnings`
- rustdoc with `-D warnings`, default and `test-support`
- independent post-edit regression review: clean

Relates to #656; does not close it.

## Summary by CodeRabbit

- **Tests**
  - Improved coverage for document resynchronization and dependent diagnostics updates.
  - Added deterministic checks to confirm fresh disk content is reflected correctly after a document closes.
  - Strengthened validation that diagnostics are cleared and resynchronization completes reliably.

PR #672, opened 2026-07-19
Simplify scope artifact finalization

jbearak:
## Summary

- make `compute_artifacts` delegate to the canonical metadata-aware builder with `metadata = None`
- append the complete scope-event timeline before one stable sort, function-scope tree rebuild, and annotation pass
- preserve the load-bearing equal-position order by appending synthetic Shiny/foreach scopes before library and `load_all()` events
- avoid building the AST source path-dedup set on the metadata-free path

## Architecture review

Fable reviewed the exact callers, invariants, and proposed tests before implementation. Its ordering amendment is reflected here: real/AST/metadata events are collected first, synthetic scopes are appended before later package effects, and only then is the complete timeline finalized once. The static-script facts path retains the same append-synthetics-then-finalize boundary.

## Tests

- `cargo test -p raven cross_file::scope::tests:: -- --nocapture` — 368 passed
- `cargo test -p raven --features test-support` — 6487 passed, 12 ignored; integration suites and 53 doctests passed
- metadata-free versus enriched-None full artifact parity across mixed ordinary events, Shiny/foreach/capture scopes, and malformed incomplete syntax
- mixed AST/metadata coverage for directive sources, ordered tar batches, declarations, NSE/standalone hash inputs, and scoped library/load-all annotations
- `cargo fmt --all --check`
- `cargo clippy --workspace --all-targets --features test-support -- -D warnings`
- `RUSTDOCFLAGS="-D warnings" cargo doc -p raven --no-deps --document-private-items`
- `RUSTDOCFLAGS="-D warnings" cargo doc -p raven --no-deps --document-private-items --features test-support`

Independent regression/new-bug review found no correctness issues after verifying stable equal-key ordering, explicit capture anchors, `source.is_function_scoped`, static-script facts, mixed metadata, hash inputs, and malformed syntax.

Refs #656

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consistency when generating cross-file scope information with or without metadata.
  * Corrected event ordering for mixed metadata and source-based timelines, including effect-position tie handling.
  * Refined handling of package-wide `load_all` sentinel package-load events and function-scope boundary behavior.

* **Tests**
  * Added/expanded coverage to ensure parity between metadata-free and enriched builders across timelines, interval data, hashes, and scope-at-position queries.
  * Strengthened validation of `load_all()` and sentinel presence/absence.

PR #671, opened 2026-07-19
Stabilize config reload and close race tests

jbearak:
## Summary

- replace shared force-marker observation in the config-reload parallel-publish test with an invocation-scoped scheduled/completed JoinSet capture
- exclude asynchronous diagnostic reservation bookkeeping from the durable close-cancellation snapshot
- isolate all three CAS/join fixtures from unrelated startup workspace-index and package-routing work

## CI evidence

This follows two Cargo Tests failures on #670:

- attempt 1: `watched_files_reload_publishes_all_open_documents_in_parallel` observed a setup-owned pre-reload force marker (`1` vs `0`)
- retry: `did_close_target_edit_is_terminal_and_never_rebases` differed only in asynchronously drained reservation/force bookkeeping (`3/1` vs `2/0`), while `did_close_retries_non_subject_watched_generation_and_leaves_unrelated_untouched` exhausted its final bounded CAS attempt after unrelated default initialization churn and left the alias open

The failing Cargo job is [here](https://github.com/jbearak/raven/actions/runs/29703387985/job/88237021147). The changes are confined to `backend.rs` and `state.rs`; the new capture and payload are `cfg(test)`, and production JoinSet output remains `()`.

Relates to #656. Does not close it.

## Validation

- three exact regression tests
- full `project_config_initialize_tests` module under 16 threads: 328 passed
- 100 rounds of all three affected tests: 300/300 passed
- two consecutive `cargo test -p raven --features test-support` runs: 6482 unit tests, all integration tests, and 53 doctests passed each run
- `cargo fmt --all --check`
- `cargo clippy --workspace --all-targets --features test-support -- -D warnings`
- rustdoc with `-D warnings`, default and `test-support`
- independent cause audit and final regression/new-bug review clean
- Fable architecture review approved before implementation

## Summary by CodeRabbit

* **Tests**
  * Improved coverage for configuration reloads triggered by watched-file changes.
  * Added validation that document updates are scheduled and completed consistently during reload publishing.
  * Expanded LSP test scenarios with controlled initialization settings and quiescent workspace behavior.
  * Improved test synchronization and deterministic result checking for reload-related workflows.

PR #686, opened 2026-07-20
Stabilize asynchronous backend test completion

jbearak:
## Summary

- give watched-file finalization a typed, invocation-owned completion receipt that survives inline retries, deferred routing, supersession, and terminal retirement
- make libpath prearm test barriers journal-scoped and await the exact blocking worker after cancellation or shutdown
- keep the production `system.file()` derivation lane process-global, while giving each unit-test `WorldState` an independent semaphore/slot lane
- replace host-dependent package setup with explicit quiescent or synthetic package fixtures, preserving real package behavior only where it is the test subject
- improve timeout diagnostics with the outstanding causal owner, target lifecycle, routing generation/slot/permit, and close-permit state

This is the bounded systemic stabilization requested after repeated Cargo failures across #669, #671, #673, and #672. It is based exactly on current `main` at `958d71e3`. It does not modify `scope.rs`.

## Failure matrix

| Failure family | Detached/background work | Previous observation boundary | Shared state/resource | Irrelevant setup work |
|---|---|---|---|---|
| #669 watched/package handoffs | watched inline retries, deferred routing finalizer, diagnostic fanout | cumulative reservation/marker deltas and eventual state | process-global routing lane; shared diagnostic bookkeeping | default package discovery, R/provider load, libpath watcher |
| #671 config reload and close races | joined diagnostic publication plus later close disk-resync/revalidation | force-marker snapshots, fixed waits, final bounded CAS outcome | shared force/pending maps; unrelated package generation churn | workspace scan and package routing in direct service fixtures |
| #673 close resync/package overlay | unjoined close resync and later diagnostic consumers | eventual unified-index/diagnostic state and an enlarged resync timeout | close permits and process-global routing slot | host package/R/libpath startup before the target operation |
| #672-era Cargo failures | watched disk read/finalizer, dirty DESCRIPTION convergence, routing retries | handler return, polling, and fixed routing deadlines | one process-global derivation permit/slot across independent test worlds | ordinary open/edit/close fixtures starting host package work |

The shared cause was not insufficient timeout length. Independent tests were starting production-default host package initialization and competing for one physical derivation lane while fixed deadlines aged before their own CAS attempts. Several assertions also treated handler return or backend-wide counters as invocation completion even though finite descendant work remained detached.

## Deterministic boundaries

- `FinalHandoffClaimLineage` records one typed `Finalized` outcome, or `RetiredBeforeFinalHandoff` when the last normal owner terminates without reaching the handoff. Its causal receipt closes only after every owned descendant completes.
- watched reopen coverage pauses after the stale disk projection, opens the buffer, releases the exact precommit barrier, and awaits the watched invocation's receipt before asserting.
- libpath prearm pauses expose exact arrival and whole-worker completion. Cancellation/shutdown tests release and await that worker without timing inference.
- package fixtures are explicit:
  - quiescent fixtures disable irrelevant workspace scan/package routing;
  - synthetic package fixtures retain package mode and the production commit/CAS path but inject one empty ready builder result;
  - real R/provider behavior remains in its dedicated suites.
- production still has one process-global derivation lane. Only unit-test worlds get independent lanes, so retries within one invocation serialize without unrelated tests spending its deadline.

No timeout was increased, no sleep or broad serialization was added, and no test-specific production workaround was introduced.

## Validation

- watched reopen + libpath prearm deterministic stress: 100/100 repetitions each
- backend project/config block at 32 threads: 331/331 passed
- complete backend block at 32 threads: 562/562 passed
- two consecutive `cargo test --workspace --all-targets --features test-support` runs: green, including integration and benchmark targets
- `cargo fmt --all --check`
- rustdoc with `-D warnings`, default and `test-support`
- two independent final reviews: async/completion review approved; fixture/new-regression review approved

## Local comparison

The exact-main 331-test project/config block exceeded five minutes, reported dozens of 60-second stalls, and produced a real failure before being stopped. The stabilized branch passed the same block twice at 32 threads in 1.58s and 1.56s.

Relates to #656. Does not close it.

### DIFF
--- a/crates/raven/src/backend.rs
+++ b/crates/raven/src/backend.rs
@@ -2042,7 +2042,25 @@ impl Backend {
             };
         }
 
-        let (library, ready, load_notes, init_error) = if force_not_ready {
+        #[cfg(test)]
+        let synthetic_build = self
+            .state
+            .write()
+            .await
+            .package_library_build_outcome_for_test
+            .take();
+        let (library, ready, load_notes, init_error) = if let Some((library, ready)) = {
+            #[cfg(test)]
+            {
+                synthetic_build
+            }
+            #[cfg(not(test))]
+            {
+                None
+            }
+        } {
+            (library, ready, Vec::new(), None)
+        } else if force_not_ready {
             (
                 Arc::new(crate::package_library::PackageLibrary::new_empty()),
                 false,
@@ -4555,7 +4573,7 @@ async fn run_system_file_convergence_transaction_owned(
             state.system_file_routing_owner_identity()
         }
     };
-    let derivation_key = {
+    let (derivation_key, derivation_lane) = {
         let Ok(state) = deadline
             .wait("system.file derivation key read", state_arc.read())
             .await
@@ -4563,7 +4581,10 @@ async fn run_system_file_convergence_transaction_owned(
             return SystemFileOwnerOutcome::Deferred;
         };
         let library = state.package_library.clone();
-        LibraryRoutingDerivationKey::convergence(&library, owner)
+        (
+            LibraryRoutingDerivationKey::convergence(&library, owner),
+            library_routing_derivation_lane(&state),
+        )
     };
     let owner_is_current = |state: &WorldState| {
         state.system_file_routing_owner_identity() == owner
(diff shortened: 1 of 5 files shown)


jbearak:
@coderabbitai review

PR #687, opened 2026-07-20
Model pacman p_load attachment semantics

### LINKED ISSUE #660: Recognize pacman::p_load() package attachments
## Motivation

`pacman::p_load()` is a common wrapper around `library()` / `require()` that accepts multiple quoted or unquoted package names and may install missing packages:

```r
pacman::p_load(dplyr, tidyr, "ggplot2")
```

Reference:

- https://trinker.github.io/pacman/p_load.html

Raven currently does not treat these arguments as package attachments, so unqualified exports may be reported as undefined and the package set may be omitted from package prefetch/freeze discovery.

## Desired behavior

For qualified `pacman::p_load()` and bare `p_load()` when `{pacman}` is attached:

- recognize statically known quoted and unquoted package arguments;
- recognize a static `char = c(...)` package vector;
- apply ordinary position-sensitive package-attachment semantics after the call;
- feed the packages through existing export lookup, missing-package diagnostics, package prefetch/freeze discovery, interface hashing, and cross-file propagation;
- respect local shadowing of `p_load`;
- handle malformed or partially dynamic calls conservatively.

## Non-goals

- Installing or updating packages.
- Executing `p_load()` or evaluating arbitrary `char` expressions.
- Inferring package names from GitHub repository strings passed to `p_load_gh()`.
- Broad support for unrelated `{pacman}` management helpers.

`p_unload()` may be assessed separately. Modeling removal would require a position-sensitive package-detach event, which is broader than the high-value attachment case and should not block this issue.

## Acceptance tests

- qualified and correctly in-scope bare calls;
- multiple symbols, strings, and static `char = c(...)`;
- dynamic/invalid arguments and a shadowed local `p_load`;
- package exports in diagnostics/completion/hover;
- missing-package behavior and CLI package discovery;
- propagation through existing explicit source relationships;
- no regression to `library()`, `require()`, or generic user functions named `p_load`.

jbearak:
Closes #660.

This models package attachment separately from namespace loading so bare pacman::p_load-style calls activate only when pacman is actually attached, while qualified p_load remains unconditional. Attachment state now follows ordered source/tar execution, Rprofile and test preambles, repeated source invocations, and conditional Shiny helper isolation consistently across recursive and streaming scope resolution.

The CLI warms and discovers conditional targets before diagnostics. Regression coverage uses only synthetic temporary workspaces and includes loadNamespace, dynamic/shadowed calls, pre-execution attachments, repeated sources, Shiny helpers, tar batches, and recursive/streaming parity.

Validation:
- cargo test --workspace --all-targets --features test-support (three final-state passes; two before bounded perf-only fast paths and one after)
- cargo fmt --all --check
- cargo clippy --workspace --all-targets --features test-support -- -D warnings
- both required rustdoc configurations
- independent correctness, regression, and performance review
- local Criterion comparison against exact main

## Summary by CodeRabbit

* **New Features**
  * Added support for recognizing package attachments through `pacman::p_load()` and eligible bare `p_load()` calls.
  * Improved package awareness across `.Rprofile`, test preambles, R Markdown, and Quarto documents.
  * Conditional package loading now reflects attachment state and execution order.

* **Bug Fixes**
  * Improved package warming, diagnostics, completion, and hover results for conditionally attached packages.
  * Prevented inactive or malformed `p_load()` calls from being reported as active package references.

* **Documentation**
  * Updated CLI and cross-file documentation to describe supported `p_load()` behavior and limitations.


jbearak:
@coderabbitai review

PR #688, opened 2026-07-20
Model deterministic package-loader loops

jbearak:
## Summary

- recognize deterministic for loops that attach literal package vectors through library() or require()
- share canonical package-load detection across diagnostics, CLI package collection, and package-state extraction
- preserve eager binding checkpoints and fail closed on control flow, helper shadowing, iterator mutation, or dynamic vectors
- document the supported pattern and architecture

## Regression coverage

All fixtures are synthetic and repository-independent. Tests cover inline and eager named vectors, trailing NULL, source-child propagation, require(), conditional and dynamic rejection, loop control flow, iterator mutation, loader shadowing, deferred function bodies, CLI package collection, and package-state extraction.

## External A/B

Against the reported checkout, exact current main produced 253 undefined-symbol diagnostics plus 2 unresolved-source diagnostics. This patch produced 150 plus the same 2 unresolved-source diagnostics: 103 false positives removed and 0 diagnostics added. The bogus iterator-package footer was also removed.

## Performance

Local Criterion comparison for artifact_compute/representative_400loc: main median 4.7321 ms; patched median 4.7752 ms (about +0.9%).

## Validation

- two complete cargo test --workspace --all-targets --features test-support runs
- cargo fmt --all --check
- cargo clippy --workspace --all-targets --features test-support -- -D warnings
- rustdoc with warnings denied, default and test-support configurations
- git diff --check
- two independent agent reviews, including iterative regression/new-bug review

## Summary by CodeRabbit

* **New Features**
  * Added detection for deterministically statically-resolved `for` loops that load packages via `library()`/`require()` patterns.
  * Improved recognition of statically defined package vectors, including `c(...)` vectors containing `NULL`.

* **Bug Fixes**
  * Fixed package visibility so packages loaded inside supported `for` loops only appear after the loop completes.
  * Refined `requireNamespace()` package extraction to avoid promoting unrelated string arguments.

* **Tests**
  * Added/extended unit tests covering loop attachment, projection consistency, and safe package-vector parsing.

* **Documentation**
  * Documented deterministic package-loader loops and updated static package-vector rules.

### DIFF
--- a/crates/raven/src/cli/packages.rs
+++ b/crates/raven/src/cli/packages.rs
@@ -758,9 +758,9 @@ fn scan_workspace_referenced_packages(root: &std::path::Path) -> Vec<String> {
 /// - the `namespace_identifier` (LHS) of each `pkg::name` / `pkg:::name`
 ///   (`namespace_operator` node — the package id is `child(0)`, confirmed against
 ///   `find_namespace_context` in `handlers.rs`), and
-/// - the first argument of each `library`/`require`/`loadNamespace`/
-///   `requireNamespace` call (mirrors `state::extract_loaded_packages`, extended
-///   to cover `requireNamespace`).
+/// - statically detected package loaders from the canonical detector, and
+/// - a string literal statically matched to `requireNamespace()`'s `package`
+///   formal.
 fn collect_referenced_packages(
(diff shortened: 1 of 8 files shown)


jbearak:
@coderabbitai review

PR #696, opened 2026-07-21
Suppress undefined cascades for packages outside active renv

### LINKED ISSUE #692: Suppress undefined-variable cascades for packages outside the active renv library
## Problem

When an explicit workspace renv is active and a referenced package is absent from the active renv library but installed in a vanilla system/user library, Raven should report the actionable `package-outside-active-library` diagnostic. It should not also produce an undefined-variable warning for every known export used from that package.

For example:

```r
library(dplyr)
filter(df, x > 0)
```

If `dplyr` exists system-wide but not in the active renv, the useful result is one package/environment warning. Reporting `filter` as undefined is cascading noise from the same root cause.

## Desired behavior

- Report `package-outside-active-library` at most once per package per document, as today.
- Use export names available from the outside installation to suppress cascading `undefined-variable` diagnostics.
- Keep availability separate from symbol knowledge: the outside installation must not make the package count as installed in the active environment.
- Do not allow the outside installation to provide normal active-environment behavior such as package loading, completion, `system.file()`, or other runtime/path-sensitive features.
- If no export metadata is available from Tier 1/2/3 or the outside installation, degraded undefined-variable diagnostics may remain, but they should not be the intended behavior.

## Rationale

The package/environment diagnostic identifies the fix: restore or install the package into the project library. Repeating an undefined-variable warning for each export adds little information and can obscure unrelated, genuine undefined names. This also follows Raven's existing design principle that export resolution and install status are separate concerns.

## Acceptance criteria

- A regression test covers a package present only in a vanilla library removed by renv activation.
- Bare exported functions following `library()` / `require()` do not receive `undefined-variable` diagnostics.
- The document still receives `package-outside-active-library`.
- Genuinely unknown symbols remain diagnosed.
- Qualified references (`pkg::member`) receive the environment warning without an undefined-variable cascade, while existing namespace-member validation semantics remain intact.

jbearak:
## Summary

- Retain positive-only explicit NAMESPACE exports in the existing diagnostic-only outside-renv overlay.
- Suppress bare undefined-variable cascades only after a true position-aware `library()` / `require()` attachment and only in the existing single-workspace case.
- Keep outside installations unavailable to normal package caches, completion, namespace resolution, loading, `system.file()`, and filesystem routing.
- Preserve genuine unknowns, pre-attachment uses, `loadNamespace()` semantics, multi-root fail-closed behavior, and S3-registration-only names.

## Validation

- `cargo fmt --all --check`
- Both required rustdoc configurations
- Two full `cargo test --workspace --all-targets --features test-support` runs
- Independent nested cause and final regression/new-bug review
- Generated end-to-end A/B fixtures with zero-package control and 500 outside packages
- Measured overhead approximately 18–28 µs per outside package (9–14 ms at deliberately large N=500); no performance blocker

Closes #692

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced false undefined-variable diagnostics after packages are attached from outside the active environment.
  * Prevented export evidence from incorrectly affecting namespace loading, multi-root sessions, S3 registrations, or later source files.
  * Ensured only explicitly declared package exports suppress relevant diagnostics.

* **Documentation**
  * Clarified how outside-library package evidence affects diagnostics, refresh behavior, and package availability.

### DIFF
--- a/crates/raven/src/handlers.rs
+++ b/crates/raven/src/handlers.rs
@@ -6374,6 +6374,26 @@ fn package_available_outside_active_renv(snapshot: &DiagnosticsSnapshot, package
             .package_available_outside_active_renv(package)
 }
 
+/// Whether a true search-path attachment supplies `name` from an installation
+/// hidden by this workspace's active renv library.
+///
+/// This is intentionally narrower than ordinary package resolution: the
+/// single-workspace gate matches the actionable environment diagnostic, and
+/// callers pass `ScopeAtPosition::attached_packages` so `loadNamespace()` and
+/// NAMESPACE imports cannot acquire bare-name attachment semantics.
+fn attached_outside_active_renv_exports(
+    snapshot: &DiagnosticsSnapshot,
+    name: &str,
+    attached_packages: &HashSet<String>,
+) -> bool {
+    snapshot.workspace_folders.len() == 1
+        && attached_packages.iter().any(|package| {
+            snapshot
+                .package_library
+                .outside_active_renv_has_explicit_export(package, name)
+        })
+}
+
 /// Emit `namespace-member-not-found` for `pkg::member` references whose package
 /// has a *complete* export set that does not contain `member`. Exports-
 /// authoritative: a `Partial`/`Unknown` set (or a never-warmed package) stays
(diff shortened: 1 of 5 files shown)


jbearak:
@coderabbitai review

PR #699, opened 2026-07-21
Make raven lint enable native linting

### LINKED ISSUE #691: `raven lint` should run lints even when editor linting is disabled
## Summary

Running `raven lint` currently honors the resolved `linting.enabled` master switch. If no `raven.toml` or configured `.lintr` is discovered, if `--no-config` is passed, or if linting is otherwise disabled, the command silently emits no lint diagnostics and exits successfully.

That behavior is surprising for an explicit `lint` command. Invoking `raven lint` expresses the user's intent to run the native linter; the editor/LSP master switch should not turn the standalone lint command into a no-op.

## Current behavior

```sh
raven lint .
```

With no lint configuration opting the project in, Raven resolves `LintConfig.enabled` to `false`. `run_lints` immediately returns an empty diagnostic list, and the command exits `0`.

`raven lint --no-config .` likewise disables linting, despite the help text saying that `--no-config` uses built-in defaults.

## Expected behavior

`raven lint` should run the native lint rules whenever the command is invoked.

- Project configuration should still control rule settings and severities.
- Per-file overrides/exclusions should still be honored.
- `--no-config` should lint using built-in defaults, as its help text implies.
- The `raven.linting.enabled` master switch should continue to govern editor/LSP lint publication, but should not disable an explicit standalone CLI lint invocation.

If users need a CLI-level way to skip linting entirely, they can simply avoid invoking `raven lint`; a successful no-op is particularly risky in CI because it looks like linting passed.

## Relevant code

- `crates/raven/src/cli/lint.rs`: resolves the ordinary lint configuration and passes it unchanged into the file walk.
- `crates/raven/src/linting/mod.rs`: `run_lints` returns immediately when `config.enabled` is false.
- `docs/cli.md`: currently documents the no-op behavior and would need updating with the fix.

jbearak:
## Summary

- make an explicit `raven lint` invocation enable Raven's native linter
- preserve configured rule values and severities
- continue honoring per-file disabled overrides and exclusions
- document the CLI-only exception to the editor/LSP master switch

## Root cause

The standalone lint command passed the ordinary resolved `linting.enabled` value into `run_lints`. With no opt-in configuration, under `--no-config`, or with a base `enabled = false`, that value was false and `run_lints` returned no diagnostics, making an explicit lint command silently succeed.

The CLI now normalizes only the merged base `linting.enabled` value before deriving the typed lint configuration and override base section. This opts the command into native linting while leaving all other rule settings and per-file overrides intact.

## Validation

- focused CLI end-to-end tests, including no-config/default activation, custom values, matching override inheritance, per-file disabled override, and rule severity `off`
- `cargo fmt --all --check`
- `RUSTDOCFLAGS='-D warnings' cargo doc -p raven --no-deps --document-private-items`
- `RUSTDOCFLAGS='-D warnings' cargo doc -p raven --no-deps --document-private-items --features test-support`
- `cargo test --workspace --all-targets --features test-support` (two complete passes)

Closes #691

## Summary by CodeRabbit

* **New Features**
  * `raven lint` now enables built-in linting rules by default, even without configuration, with `--no-config`, or when base linting is disabled.
  * Existing per-file exclusions, rule severities, and overrides continue to apply.

* **Bug Fixes**
  * Rules explicitly set to `off` remain silent when native linting is enabled.

* **Documentation**
  * Updated CLI and linting guides to explain the behavior and its scope.


jbearak:
@coderabbitai review

